PR #8, opened 2018-08-15
Expose error message when .install.sh fails

TengHu:
1) Expose error message when .install.sh fails with subprocess.run (Python 3.5+)
2) Introduce binding version and h3 library version (3.1.0)
3) Upgrade binding version to 3.0.1

### DIFF
--- a/.install.sh
+++ b/.install.sh
@@ -15,7 +15,7 @@ git clone https://github.com/uber/h3.git h3c
 
 pushd h3c
 git pull origin master --tags
-git checkout "v$VERSION"
+git checkout "$VERSION"
 
 # Run CMake, installing a recent version if not found or not compatible
 {
--- a/binding_version.py
+++ b/binding_version.py
@@ -1,1 +1,1 @@
-binding_version = '3.0.1'
+binding_version = '3.1.0'


binding_version.py
```diff
@@ -0,0 +1 @@
+binding_version = '3.0.1'
```
nrabinowitz: Please see the [versioning guidelines](https://github.com/uber/h3/blob/master/dev-docs/creating_bindings.md#documentation) ([JS version](https://github.com/uber/h3-js/blob/master/README.md#versioning)). After much debate, we decided that bindings should keep major/minor in sync with the core library, and maintain their own patch version.

.install.sh
```diff
@@ -2,12 +2,20 @@
...
+    exit 1
+fi
+
 mkdir -p h3/out
 rm -rf h3c
 git clone https://github.com/uber/h3.git h3c
 
 pushd h3c
 git pull origin master --tags
+git checkout "v$VERSION"
```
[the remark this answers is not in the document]
TengHu: done

binding_version.py
TengHu: upgrad binding version to 3.1.0

PR #15, opened 2018-10-31
Add H3 version badge

isaacbrodsky:


isaacbrodsky:
Build is failing because the build script is trying to pull a more recent version of the H3 core library. I'll update this PR to pull a specific version, so the CI test is consistent, and then make a separate PR for updating to H3 3.2.0.

PR #25, opened 2019-02-06
Fix subprocess api so that the install is backward compatible with python 2.7

TengHu:
Fix subprocess api so that the install is backward compatible with python 2.7


karentycoon:
Can someone grant me write access? Thanks!

TengHu:
Done

PR #26, opened 2019-02-21
Upgrade to 3.4.1

TengHu:
1) Upgrade to 3.4.1
2) Fix h3_set_to_multi_polygon unit tests
3) reformat tests/test_h3.py

### DIFF
--- a/tests/test_h3.py
+++ b/tests/test_h3.py
@@ -22,7 +22,7 @@ from h3 import h3
 
 class TestH3Core(unittest.TestCase):
     def shift_circular_list(self, start_element, elements_list):
-        # We shift the circular list so that it starts from start_coord,
+        # We shift the circular list so that it starts from start_element,
         start_index = elements_list.index(start_element)
         return elements_list[start_index:] + elements_list[:start_index]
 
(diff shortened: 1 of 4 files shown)


tests/test_h3.py
```diff
@@ -197,190 +231,301 @@ def test_polyfill_bogus_geo_json(self):
...
+                [
+                    vertices[2], vertices[3], vertices[4], vertices[5],
+                    vertices[0], vertices[1]
+                ]
+            ]
+        ]
+
+        # We shift the multi_polygon circular list so that it starts from expected[0][0][0],
+        # since output starting from any vertex would be correct as long as it's in order.
+        multi_polygon[0][0] = self.shift_circular_list(
```
nrabinowitz: It seems like it would be a lot easier just to manually change the order in `expected`? In general I try to avoid extra data transformations in tests, as it makes them harder to debug.

It might actually feel more valid here to try every starting node, and assert that at least one was correct (closer to property-based testing). Either way, this is up to you - I think it's okay like this if you prefer.

TengHu: That makes sense. The problem I found is once the core algorithm changes, the order of polygon coords may change which makes h3_set_to_multi_polygon tests very brittle.

Yes, I will find a way to configure the code formatter since extraneous formatting change like this is really distracting.

TengHu:
> I presume the new features in the updated H3 will come later?

@dfellis Yep :)

PR #29, opened 2019-03-01
Add h3Line support

wwchen:


h3/h3.py
```diff
@@ -754,3 +757,11 @@ def get_h3_unidirectional_edge_boundary(h3_address, geo_json=False):
 def h3_distance(h3_address_origin, h3_address_h3):
     return libh3.h3Distance(
         string_to_h3(h3_address_origin), string_to_h3(h3_address_h3))
+
+
+def h3_line(h3_address_origin, h3_address_h3):
+    array_len = h3_distance(h3_address_origin, h3_address_h3) + 1
```
isaacbrodsky: `h3LineSize` is a better choice here, because you can also check for sizes <0 which indicate the library cannot compute the line.

h3/h3.py
```diff
@@ -754,3 +757,11 @@ def get_h3_unidirectional_edge_boundary(h3_address, geo_json=False):
...
+    IndexArray = c_long * array_len
+    index_array = IndexArray()
+    libh3.h3Line(string_to_h3(h3_address_origin), string_to_h3(h3_address_h3), index_array)
```
isaacbrodsky: This ignores a possible error return code from `h3Line`

PR #31, opened 2019-03-05
upgrade version to 3.4.2

karentycoon:


karentycoon:
CLA is already signed. Not sure why it is showing unsign error.

PR #35, opened 2019-03-13
Show missing dependency error to user for all OS types

karentycoon:
The linux version of cmake does not work on OSX environment. So adding a machine type check here, if machine type is Darwin, then don't install cmake but return error to user.

Not sure how Windows platform should work so not checking windows platform.

Tested by uninstall cmake and run install script. It will show error message
`Failed to find cmake. Please make sure cmake is installed and set in PATH.`

### DIFF
--- a/.install.sh
+++ b/.install.sh
@@ -9,6 +9,10 @@ if [ "" == "$VERSION" ]; then
     exit 1
 fi
 
+command -v cmake  >/dev/null 2>&1 || { echo "cmake required but not found."; exit 1; }
+command -v make  >/dev/null 2>&1 || { echo "make required but not found."; exit 1; }
+command -v cc  >/dev/null 2>&1 || { echo "cc required but not found."; exit 1; }
+
 mkdir -p h3/out
 rm -rf h3c
 git clone https://github.com/uber/h3.git h3c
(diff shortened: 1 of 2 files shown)


.install.sh
```diff
@@ -21,15 +21,21 @@ git checkout "$VERSION"
...
+          mkdir cmake-download
+          pushd cmake-download
+          curl -O https://cmake.org/files/v3.10/cmake-3.10.0-rc5-Linux-x86_64.sh
+          bash cmake-3.10.0-rc5-Linux-x86_64.sh --skip-license
+          export PATH=`pwd`/bin:$PATH
+          echo $PATH
+          popd
+          cmake -DENABLE_FORMAT=OFF -DBUILD_SHARED_LIBS=ON .
+	else
+	  echo "Failed to find cmake. Please make sure cmake is installed and set in PATH."
```
dfellis: I'd personally prefer it just do this for all operating systems.

isaacbrodsky: +1, this code does not appear it would work correctly for x86 Linux, or ARM Linux. My understanding is there is not a technical need for the installation process to pull in a CMake version, so this is something we should leave to the user to supply.

karentycoon: I agree. It's also not perfect to modify user's PATH variable. Updated the code change.

Issue #32, opened 2019-03-08
Installation on Windows 10 64 bit


shahroze90:
I have tried to install h3 on windows 10 64 bit with python 3.6.
I ran the command pip install h3.
After install, when i run the program, i get this error:
```
    from h3 import h3
  File "C:\Users\Shahroze Tariq\AppData\Local\Programs\Python\Python36\lib\site-packages\h3\h3.py", line 39, in <module>
    libh3 = cdll.LoadLibrary('{}/{}'.format(_dirname, libh3_path))
  File "C:\Users\Shahroze Tariq\AppData\Local\Programs\Python\Python36\lib\ctypes\__init__.py", line 426, in LoadLibrary
    return self._dlltype(name)
  File "C:\Users\Shahroze Tariq\AppData\Local\Programs\Python\Python36\lib\ctypes\__init__.py", line 348, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: [WinError 126] The specified module could not be found
```

karentycoon:
Could you clear your pip cache, make sure cmake is installed and try again?

dfellis:
@karentycoon h3-py does not work on Windows, period, right now. [`h3/h3.py` has OS-detection code that ignores Windows](https://github.com/uber/h3-py/blob/master/h3/h3.py#L37)

[I have started work on getting h3-py to work on Windows](https://github.com/dfellis/h3-py/pull/1) but I can only do it in my spare time since I don't work at Uber anymore, and I also don't have any Windows computers at my house, so I'm relying on Appveyor for debugging what's wrong. It's now at the point where a proper DLL is being generated, but I can't get Python on Windows to load it, yet.

karentycoon:
Got it. Thanks for your detailed explain @dfellis ! I am new here to mange this repo. I have a windows computer at my home. If you wanna try out some tests, I'd love to help!
Could we merge all windows installation related issue to one issue for better tracking?

shahroze90:
I can see some commits related to windows build #37 and #46 . However, this repo still does not work on windows

dfellis:
@shahroze90 can you provide more detail? [h3-py is being tested on 32 and 64bit Windows for Python 2 and 3](https://ci.appveyor.com/project/Uber/h3-py/branch/master).

So without a build log we cannot help you.

shahroze80:
PS D:\Documents\test> pip3 install h3
Collecting h3
[...]

PR #37, opened 2019-03-15
Windows build

isaacbrodsky:
Addresses #32

This PR is based on @dfellis' branch working on Windows support and includes his commits too.

Adds ability to build on Windows, which was tested on Appveyor. We'll need to enable Appveyor on this repository, too.

There are a number of assumptions built into the build script which probably need to be documented. Specifically, if you want a 64-bit build, you must specify the environment variable `PYTHON_ARCH=64`. I'm not sure what the more idiomatic way of doing this is, perhaps we should check the processor architecture env variable?

Several issues were fixed to get it working on Windows:
* The type of `H3Index` should be `c_ulonglong`. On Windows, `long` and `c_long` are always 32 bits. I assigned this its own name in h3.py to be more explicit about what we mean.
* Fixed a test that was comparing floats for equality.
* Python 2.7: Fixed a test that checked for a number being an `int`. I presume it was a `long` instead.
* Python 2.7: Looks to me like formatting of indexes as strings was not working in Python 2.7 because it added a 'L' suffix.

### DIFF
--- a/.install.sh
+++ b/.install.sh
@@ -43,9 +43,9 @@ if command -v make; then
   fi
 else
   if [[ "${Platform}" == "x64" ]]; then
-    cmake . -DCMAKE_C_FLAGS='"/MT"' -DCMAKE_VS_PLATFORM_NAME=${Platform} -G "Visual Studio 14 Win64" && msbuild.exe h3.vcxproj;
+    cmake . -DCMAKE_VS_PLATFORM_NAME=${Platform} -G "Visual Studio 14 Win64" && msbuild.exe h3.vcxproj;
   else
-    cmake . -DCMAKE_C_FLAGS='"/MT"' -DCMAKE_VS_PLATFORM_NAME=${Platform} && msbuild.exe h3.vcxproj;
+    cmake . -DCMAKE_VS_PLATFORM_NAME=${Platform} && msbuild.exe h3.vcxproj;
   fi
   ls -l bin/Release/*
   cp bin/Release/h3.lib ../h3/out
--- a/.install.sh
+++ b/.install.sh
@@ -43,9 +43,9 @@ if command -v make; then
   fi
 else
   if [[ "${Platform}" == "x64" ]]; then
-    cmake . -DCMAKE_C_FLAGS='/MT' -DCMAKE_VS_PLATFORM_NAME=${Platform} -G "Visual Studio 14 Win64" && msbuild.exe h3.vcxproj;
+    cmake . -DCMAKE_C_FLAGS='\/MT' -DCMAKE_VS_PLATFORM_NAME=${Platform} -G "Visual Studio 14 Win64" && msbuild.exe h3.vcxproj;
   else
-    cmake . -DCMAKE_C_FLAGS='/MT' -DCMAKE_VS_PLATFORM_NAME=${Platform} && msbuild.exe h3.vcxproj;
+    cmake . -DCMAKE_C_FLAGS='\/MT' -DCMAKE_VS_PLATFORM_NAME=${Platform} && msbuild.exe h3.vcxproj;
   fi
   ls -l bin/Release/*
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -54,7 +54,9 @@ def __init__(self, attrs=None):
     },
     package_data={
         'h-py':
-        ['out/*.dylib' if platform.system() == 'Darwin' else 'out/*.so.*']
+        ['out/*.dylib' if platform.system() == 'Darwin' else (
+            'out/*.dll' if platform.system() == 'Windows' else
```
dfellis: Now that I'm looking at this as a reviewer, I wonder if this should be `'out\\*.dll'` here to match how `libh3_path` is defined in `h3/h3.py` above?

isaacbrodsky: Updated, but I standardized on forward slashes in `h3/h3.py` instead.

dfellis: Does that work? I had the backslashes before because it wasn't concatenating the path correctly in Appveyor.

isaacbrodsky: Appears to be working in Appveyor: https://ci.appveyor.com/project/isaacbrodsky/h3-py/builds/23110297

karentycoon:
I've tried this patch in my windows machine, but got following errors:
`--Selecting Windows SDK version to target Windows 10.0.17134.
Cmake Error at
AMakeLists.txt:54 (project): Failed to run MSBuild command:
MSBuild.exe
to get the value of VCTargetsPath`
I've installed Visual Studio and powershelgl. Any steps I'm missing here?

isaacbrodsky:
Which versions of Visual Studio, Windows, Python, and which architecture? I suspect this has to do with the script assuming Visual Studio 2015.

.install.sh
```diff
@@ -21,14 +19,29 @@ pushd h3c
...
+  make
+  ls -l lib/libh3*
+  cp lib/libh3* ../h3/out
+  if [ -e ../build ] && [ -d ../build ]; then
+      LIBNAME=`ls ../build/ | grep '^lib'`
+      mkdir -p ../build/$LIBNAME/h3/out
+      cp lib/libh3* ../build/$LIBNAME/h3/out
+  fi
+else
+  # Assumed to be Windows, default to x64
```
karentycoon: I'm bit worried maybe some users have installed make in their windows environment? Maybe we can use uname to be more safe here?

karentycoon:
> Which versions of Visual Studio, Windows, Python, and which architecture? I suspect this has to do with the script assuming Visual Studio 2015.

karentycoon:
> Which versions of Visual Studio, Windows, Python, and which architecture? I suspect this has to do with the script assuming Visual Studio 2015.

It's my environment's problem. I'll try again.

isaacbrodsky:
There's also beginning of a Cython binding, which may replace this approach: https://github.com/isaacbrodsky/h3-py/tree/cython I haven't tested yet if the Cython branch builds on Windows, but I would expect that it does.

.install.sh
isaacbrodsky: I'm not sure what `uname` to look for. Maybe anything starting with `MINGW`? On my system it includes the Windows version as part of the OS name.

Another option might be to change from specifying the generator to specifying the architecture with `-A`, in that case we might be able to use CMake to invoke the build system, regardless of which build system is used.

isaacbrodsky: @karentycoon Updated, it now uses the `OS` environment variable to detect Windows.

andrewong11:
Hi, I am quite new with programming. may I know if h3 can now be properly installed on windows? I have tried pip install h3, but it seems that the version of h3 installed is the one without these windows updates. I've also tried cloning the github repository, but i still couldn't import the h3 package. Or could you describe how I should install h3 with vs2017? thank you very much

isaacbrodsky:
@andrewong11 I think a new version of H3-py needs to be published before it will be installable on Windows.

Issue #42, opened 2019-04-12
Bug in `k_ring_distances()`


kwmsmith:
In the vicinity of the pentagon `870800000ffffff`:

```python
>>> h3.k_ring_distances('870800003ffffff', 2)
[{'0', '870800003ffffff'},
 {'870800000ffffff',
  ...},
 {'870800004ffffff',
  ...}]
```

Note the extra null hex in the 0th set above.

Doesn't show up in `h3.k_ring('870800003ffffff', 2)` or when `k < 2`.

nrabinowitz:
This is expected output from the C core; the bindings should be responsible for removing nulls (or I suppose explicitly passing this responsibility to their consumers).

kwmsmith:
Right, in all other API functions the python wrapper removes nulls, looks like null pruning was missed for this case.

karentycoon:
@kwmsmith Thanks for reporting! We just released version 3.4.3, and this bug is fixed in the new version.

PR #44, opened 2019-04-17
remove null values in k_ring_distances and hex_range_distance

karentycoon:
#42 
hexagon_c_array_to_set has removed zero values in the output of k_ring and hex_range method.
k_ring_distances and hex_range_distances should return a similar result with 0 value excluded. 

Tested by running locally
```
>>> h3.k_ring_distances('870800003ffffff', 2)
[set(['870800003ffffff']), set(['870800002ffffff', '870800000ffffff', '87080001cffffff', '87080001dffffff', '870800005ffffff', '87080002affffff']), set(['870800018ffffff', '87080002bffffff', '870800019ffffff', '87080002effffff', '870800015ffffff', '870800011ffffff', '870800028ffffff', '87080001effffff', '870800006ffffff', '870800004ffffff', '870800023ffffff'])]
>>> h3.k_ring('870800003ffffff', 2)
set(['870800018ffffff', '87080002bffffff', '870800019ffffff', '870800003ffffff', '87080002effffff', '870800015ffffff', '870800011ffffff', '870800000ffffff', '87080001cffffff', '870800002ffffff', '870800028ffffff', '87080001dffffff', '87080001effffff', '870800005ffffff', '870800006ffffff', '87080002affffff', '870800004ffffff', '870800023ffffff'])
```

### DIFF
--- a/h3/h3.py
+++ b/h3/h3.py
@@ -601,9 +601,8 @@ def hex_range_distances(h3_address, ring_size):
     for i in range(0, ring_size + 1):
         out.append(set([]))
     for i in range(0, array_len):
-        if krings[i] != 0:
-            ring_index = distances[i]
-            out[ring_index].add(h3_to_string(krings[i]))
+        ring_index = distances[i]
+        out[ring_index].add(h3_to_string(krings[i]))
     return out
 
 
--- a/h3/h3.py
+++ b/h3/h3.py
@@ -352,8 +352,9 @@ def k_ring_distances(h3_address, ring_size):
     for i in range(0, ring_size + 1):
         out.append(set([]))
     for i in range(0, array_len):
-        ring_index = distances[i]
-        out[ring_index].add(h3_to_string(krings[i]))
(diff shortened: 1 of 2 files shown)


karentycoon:
> Change looks good, could you add test cases for these?

Added new test case.

h3/h3.py
```diff
@@ -600,8 +601,9 @@ def hex_range_distances(h3_address, ring_size):
     for i in range(0, ring_size + 1):
         out.append(set([]))
     for i in range(0, array_len):
-        ring_index = distances[i]
-        out[ring_index].add(h3_to_string(krings[i]))
+        if krings[i] != 0:
```
isaacbrodsky: I don't believe this is needed because hexRangeDistances would return a failure code instead.

karentycoon: > I don't believe this is needed because hexRangeDistances would return a failure code instead.

I see. revert hex_range_distances change.

PR #51, opened 2019-06-13
Multiple interfaces

ajfriend:
Moving this PR over from isaacbrodsky/h3-py#5.


kwmsmith:
Lots of comments; collating here in one place rather than scattering throughout the PR:

## Module design

How do you want to break up the cython modules?
    
I could see something like this at first cut:

* Everything without an `_int()` suffix is put in the top -level `h3` api and these functions would call the relevant `h3.int_api` functions after converting their inputs and appropriate error checking, etc.  This is meant as the top-level, public interface, and it assumes the incoming data could be invalid.

* Everything that has an `_int()` suffix is grouped in a module like `h3.int_api` (modulo better name) and has the `_int()` suffix removed since it would be redundant.

* Utility stuff and `HexMem` cpdef class put into their own small modules and used where necessary.

## Scalar functions

I like having a high-level string-in / string-out interface that calls into the low-level int-in / int-out interface.
    
What about validation checks -- in practice, my expectation is that H3 would raise if passed in an invalid hex string and I'm always surprised when H3 doesn't catch an invalid hex string.  The string scalar functions seem like a natural place for that sort of thing, and the int scalar functions could forego checking for performance.

## Set / array returning functions:

I feel like having too many options as part of the public API--sets of ints, sets of strings, arrays of ints, and arrays of strings--isn't the best design.  I'd recommend providing just two interfaces, mirroring the scalar cases above: a low-level, fast and memory-efficient array-of-hex-ints interface--that lives in `h3.int_api`--and a high-level set-of-hex-strings interface that lives in the top-level API.

If users really want arrays of strings, they can call `list(hex_set)`, 'np.fromiter(hex_set)`, or `[int2hex(h) for h in array_int]` explicitly.


## HexMem

* To be strictly correct, should `n` be a `size_t`?
* Hygiene: in __dealloc__(), set `self.ptr = NULL` after calling `stdlib.free(self.ptr)`
* Bounds checks for `n` argument in `__cinit__` and `resize`
* Non-zero checks on all `n` arguments to avoid nasty edge cases with `malloc(int, 0)`
* I feel like we need to test the hell out of this class, preferably with something like hypothesis to exercise a ton of edge cases.
* recommend removing array_str method.  Side note: as implemented, it creates a numpy array of numpy strings, which is a pain to work with.  

### HexMem.memview()

regarding comment: we could ensure that `drop_zeros()` is called only when necessary by doing the following:

* rename `self.ptr` to `self._ptr` and make it private.
* add an accessor method: `cdef H3Int* get_ptr(self)`
* when accessor is called, set a private `self._dirty` bit.
* `memview()` and other public API functions modified:

        if self._dirty:
            self.drop_zeros()
            self._dirty = False
        ...

This way drop_zeros is called once when the dirty bit is set, and `dirty` is set only when `get_ptr()` is called.

### HexMem.array_int()

    def array_int(self):
        """ currently, this method copies. ideally, we'd re-use the memory
        and make sure it doens't get freed when HexMem gets garbage collected
        """

        # np.asarray does not copy, but there's no way to give it ownership of the memory...
        # np.array allocates new memory
        return np.array(self.memview())

According to the comment, "...there's no way to give [the array] ownership of the memory [with np.asarray]...".  Have you verified that using asarray here leads to segfaults or similar?
    
I don't think the comment is accurate -- using `np.asarray()` should set the array's `.base` readonly attribute to the memoryview object and hold a reference to it, preventing its deletion as long as the numpy array exists.

ajfriend:
# Module design

Totally agree that it makes sense to limit the number of options for the public API. I'll list out 4 below (modulo naming!), and of those, we should definitely have `h3.api.set_str` and `h3.api.array_int`, but I go back and forth as to weather we'd want `h3.api.set_int`. As you said, `h3.api.array_str` probably isn't necessary.

However, I don't think we need to make a binding decision right now. Those APIs should be fairly lightweight, and so we can add/delete them later.

Also, as you said, we would drop any function suffixes like `_int`, since we get that information from the module name. I only had the suffixes in this PR because I couldn't figure out how to get CMake to work with multiple Cython files :/

@kwmsmith, could you help me get started with breaking these into submodules? I tried writing separate `.pyx` files, but couldn't get CMake to get the compilation correct.

I considered an alternative design which would have everything be a wrapper around `h3.api.array_int`, but that would have the downside of locking us into a `numpy` dependency. Instead, I'm proposing wrapping each API around the `HexMem` interface.

## Internal utilities

- `h3.util.hexmem`
    - contains`HexMem` class and other utility functions
    - intended for internal use
    - ideally, **no** `numpy` dependency, to maintain flexibility

## Public APIs

Each of these should be a light wrapper around the functionality in `h3.util.hexmem`.

- `h3.api.set_str`
    + `set` of `str`s
    + should match the current `h3-py` API
    + will probably alias this to be the "default" API
    + provides error checking on inputs
- `h3.api.set_int`
    + `set` of `int`s
    + provides error checking on inputs
- `h3.api.array_int`
    + `numpy.array` of `int64`s
    + no memory copies (for speed)
    + no error checking on inputs (for speed)
- `h3.api.array_str`
    + `numpy.array` of `str`s
    + error checking on inputs?
    + of the 4 APIs, this one is probably the least-necessary

# HexMem

I agree with all your suggestions here, including that we should test the hell out of HexMem. I also like the idea of the dirty bit.

I'll try to add these changes to the PR in the next few days.

# Action items

1. I'll add changes based on your suggestions above to the PR.
2. Could you help me figure out how to get multiple `.pyx` files working so that we can modularize?

kwmsmith:
Spent time with modularizing the Cython code, with some success getting CMake to behave.  Will get a PR together so you can see the pattern.  CMake can be very frustrating, especially when combining with scikit-build and using a git submodule.

The biggest outstanding issue is that I can't figure out how to have the Python package named "h3".  Currently I renamed it to "h3py" to get things working.  This wasn't an issue when we had a monolithic extension module, but now that we want to break it up into multiple modules and they need to import/cimport from each other, they need to know their fully qualified names.  Python extension modules have very strict requirements about knowing their fully qualified names at *compile time*, and I haven't figured out a way to get around this when modularizing.

FYI, I will be on vacation starting tomorrow (Saturday) through most of next week, so I won't be able to respond to anything in the near term.

ajfriend:
Sounds good! If you figure out the pattern, I can duplicate it to finish the modularization. Enjoy vacation!

kwmsmith:
xref: https://github.com/ajfriend/h3-py/pull/1

ajfriend:
@kwmsmith I made some progress breaking things up and playing with the interfaces. Still a work in progress tho...

ajfriend:
@kwmsmith, I could use some help with the following.

I think I'm starting to better understand how to communicate memory ownership so that when the last view of the memory contained in a `HexMem` object is garbage collected, the `HexMem.__dealloc__` properly frees the memory.

I didn't fully understand that before, which is why the current implementation in `h3core` returns `HexMem` objects **themselves**, instead of memoryviews. As it stands, if I just return a memoryview of the memory owned by the `HexMem` object (and that object is GC'd), the memoryview still exists, but points to invalid freed memory.

I think I can clean up the current `h3core` code (to provide a pure memoryview interface) if I knew how to let the memoryview know that the `HexMem` object owns and should deallocate the memory. I guess this requires a change to the `HexMem.memview()` function. I looked through the documentation, but couldn't quite figure out how to do it. @kwmsmith, do you know how?

Understanding all this, I now better understand your comment above:

> I don't think the comment is accurate -- using np.asarray() should set the array's .base readonly attribute to the memoryview object and hold a reference to it, preventing its deletion as long as the numpy array exists.

Once we get this working, I think the no-copy "array of uint64s" interface should fall into place.

ajfriend:
New approach.

The trouble I was having above was due to the fact that `HexMem` was the "terminal" object that owned the memory and was responsible for freeing it. Instead, I think it makes sense to have a `cython.view.array` object as the owner and set its `callback_free_data` attribute to clean up the memory. This way, ownership is naturally managed and cleaned up when the last reference to the data is garbage collected. 

I removed the `HexMem` class and changed the "base" interface to input and output memoryview objects. The three different user-facing interfaces are just normal Python on top of the memoryview interface. (I'm guessing this is closer to what you originally had in mind, @kwmsmith.)

I've got more cleaning up to do, but I think we've finally hit on the right module design.

ajfriend:
Actually, we may want to validate **all** inputs, even for the `numpy` interface. There are some input hexagons that seem to cause segfaults (on my Cython code). For example, if `h = '8a28308280fffff'`, then `h3.k_ring(h, 1)` will segfault if we don't check the validity of the hex.

This Cython branch correctly identifies `h` as an invalid hexagon with `is_valid()`, but this doesn't seem to be the case in the other Python wrapper (see #53).

Some functions work without failing (like `h3_to_geo()`), but give unexpected results, like 

```
>>> h = '8a28308280fffff'
>>> h3.geo_to_h3(*h3.h3_to_geo(h), res=10) == h
False
```

@kwmsmith, in light of the possibility of segfaults, I wonder if it would make sense to validate **all** inputs at the Cython (uint64) level, and raise exceptions for invalid data. We were discussing avoiding the checking for the `numpy` interface for speed, but I think it might be necessary if segfaults happen.

A nice benefit of this is that if we do validate all inputs, our interface code should become a bit simpler.

ajfriend:
I feel like I've got this mostly to where I'd like it. Let me know if anyone has any thoughts!

h3py/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+add_library(util MODULE ${util} ${LIB_SOURCE_FILES} ${CONFIGURED_API_HEADER})
+python_extension_module(util)
+set_property(TARGET util PROPERTY C_STANDARD 99)
+target_compile_definitions(util PUBLIC $<TARGET_PROPERTY:h3,COMPILE_DEFINITIONS>)
+target_include_directories(util PUBLIC $<TARGET_PROPERTY:h3,INCLUDE_DIRECTORIES>)
+
+install(TARGETS h3core util LIBRARY DESTINATION h3py)
+
+
+add_cython_target(geo C PY3)
```
isaacbrodsky: Each of `geo` and `util` are the different `pyx` files? It might be good to clarify what the dependency of this block is, and extracting it to a macro to deduplicate the CMake code.

h3py/api/_api_template.py
```diff
@@ -0,0 +1,228 @@
...
+    def num_hexagons(resolution):
+        return h3core.num_hexagons(resolution)
+
+    def hex_area(resolution, unit='km'):
+        return h3core.hex_area(resolution, unit)
+
+    def edge_length(resolution, unit='km'):
+        return h3core.edge_length(resolution, unit)
+
+    def is_cell(h):
```
isaacbrodsky: Do you want to name this `h3_is_valid` to be consistent with the C API?

h3py/api/array_int.py
```diff
@@ -0,0 +1,20 @@
...
+    _out_scalar = _id,
+    _in_collection = _id,
+    _out_collection = np.asarray,
+)
+
+# todo: not sure if this is the best way to do this...
+# if too weird, we can always fall back to just cut-and-pasting the contents
+# of the `_api_functions` body. However, that isn't very DRY.
+# Something like a python #include macro would be nice here...
+globals().update(_funcs)
```
isaacbrodsky: Does this break the ability to `from ...array_int import h3_to_geo`?

h3py/api/old.py
```diff
@@ -0,0 +1,60 @@
+from .set_str import (
```
isaacbrodsky: It seems like this is an API more consistent with the C API - why is this the "old" API? Do we want to have a single API or multiple APIs? Which one should use a user of the library use?

h3py/h3core.pyx
```diff
@@ -0,0 +1,274 @@
...
+    """ compute the hex-distance between two hexagons
+    """
+    u.check_addr(h1)
+    u.check_addr(h2)
+
+    d = h3c.h3Distance(h1,h2)
+
+    return d
+
+cpdef H3int[:] disk(H3int h, int k):
```
isaacbrodsky: I'm a little concerned we're going to proliferate more terms, when we probably have too many terms for kRing already (I'd like to reduce the number of unique terms we use for that in H3 4.x)

h3py/h3core.pyx
```diff
@@ -0,0 +1,274 @@
...
+    mv = create_mv(ptr, N)
+
+    if flag != 0:
+        raise H3ValueError('Could not uncompact set of hexagons!')
+
+    return mv
+
+
+
+# weird return type here
```
isaacbrodsky: Just use `int64_t` directly here rather than H3int? The type in the C library is signed (unlike H3int)

h3py/h3core.pyx
```diff
@@ -0,0 +1,274 @@
...
+
+
+cpdef double hex_area(int resolution, unit='km') except -1:
+    u.check_res(resolution)
+
+    area = h3c.hexAreaKm2(resolution)
+
+    # todo: multiple units
+    convert = {
+        'km': 1.0,
```
isaacbrodsky: km2, not km

h3py/h3core.pyx
```diff
@@ -0,0 +1,274 @@
...
+
+    return h3c.h3IndexesAreNeighbors(h1, h2) == 1
+
+
+cpdef H3int edge(H3int origin, H3int destination) except 1:
+    u.check_addr(origin)
+    u.check_addr(destination)
+
+    if h3c.h3IndexesAreNeighbors(origin, destination) != 1:
+        raise H3ValueError('Hexes are not neighbors: {} and {}'.format(origin, destination))
```
isaacbrodsky: "Indexes are not neighbors" (one might be a pentagon)

h3py/h3core.pyx
```diff
@@ -0,0 +1,274 @@
...
+    u.check_edge(e)
+
+    return h3c.getOriginH3IndexFromUnidirectionalEdge(e)
+
+cpdef H3int edge_destination(H3int e) except 1:
+    u.check_edge(e)
+
+    return h3c.getDestinationH3IndexFromUnidirectionalEdge(e)
+
+cpdef (H3int, H3int) edge_hexes(H3int e) except *:
```
isaacbrodsky: `edge_indexes` is likely a better term here, as with above, when there's a possibility of pentagons we should prefer the term `index` for the representation of the cell.

h3py/util.pyx
```diff
@@ -0,0 +1,135 @@
...
+        raise InvalidH3Resolution(res)
+
+
+
+## todo: can i turn these two into a context manager?
+cdef H3int* create_ptr(size_t n) except *:
+    if n <= 0:
+        # raise H3ValueError("Can't allocate an array of length 0!")
+        # actually, n=0 should be OK.
+        pass
```
isaacbrodsky: What does it return if `n == 0`? Seems like it should just pass it along to `calloc`

h3py/api/array_int.py
ajfriend: This doesn't break it. That import style still works.

ajfriend:
While working through this, I had an opportunity to go through the API pretty intimately. I tried to take note of and think about any sticky points which weren't immediately obvious or were confusing (and I would imagine would be even more confusing to new users). I'll try to write down some of the major points here, with the hope that it'll be useful in the short or long term as we think about the API design. 


- When we say "H3 Index", do we mean a `uint64`, or do we mean the hex string? Maybe it isn't really necessary to make the distinction, but I would say that the majority of new users are exposed to the string interface and don't even realize the index is representable as a 64-bit integer.

- When does "hexagon" really mean "hexagon", and when does it mean "hexagon or pentagon"? Is there a better term, like "cell" or "index"? But when does "index" mean "hexagon or pentagon", and when does it mean "hexagon, pentagon, or edge"?
    - The documentation doesn't make clear which meaning is intended for `h3IsValid()`. This could be confusing for a user when contrasting to `h3IsPentagon()`.
    - Similarly for `h3IndexesAreNeighbors`, @isaacbrodsky made a comment about using "indexes" instead of "hexes". I'd love to use precise wording, but I don't think this is done everywhere in the API. For example, `getH3UnidirectionalEdgesFromHexagon()` takes in hexagons or pentagons.

- I've never met a person who can keep `k_ring` and `hex_ring` straight. In my experimental interface above, I tried to use `disk` and `ring` to refer to the "solid" and "hollow" concepts. I'm not expecting we'll make this change in the short term, but maybe something to think about for H3 4.x?


Particularly for the hexagon/index/cell issue above, I wonder if we could define a simple glossary to define each of these terms **exactly**, and then make sure that the API conforms.

ajfriend:
@isaacbrodsky @nrabinowitz @dfellis Would a separate "H3 glossary" issue make sense here?

h3py/CMakeLists.txt
ajfriend: Totally agree. I just cut and paste here. A macro would be great. Does anyone know how to do that in CMake?

h3py/api/_api_template.py
ajfriend: See my comments below, but I was worried that `h3_is_valid` might not be clear to users that it **does not** validate an edge, but that it **does** validate hexagons and pentagons. I was trying out `cell` experimentally. Overall, names are cosmetic and I'm happy to move them around. I just wanted to take a chance to try out different names in the hopes of making the API easier to understand.

h3py/api/old.py
ajfriend: I just wanted to show that it was easy to make an API which conforms to the existing `h3-py` API. The term "old" just reveals my bias as towards which API I like better ;)

We were hoping to have multiple APIs, but with one default API which would probably be under the `h3.*` top-level namespace. Other APIs might live under something like `h3.api.*`.

h3py/h3core.pyx
ajfriend: Agreed. This was just an experiment to try to fix the confusion around `k_ring` and `hex_ring`.

h3py/h3core.pyx
ajfriend: Agreed. Although I would love to see an "H3 Glossary" where all of these terms are defined unambiguously (see my comment below.)

dfellis:
The API grew organically over time as the need for certain operations arose. Some of the names, like k-ring, came from internal literature at Uber and didn't match the DDG origins of this implementation (as I recall for k-ring, it came about when an early prototype projected to an IJ coordinate system and so "K" was an available variable name to distinguish the ring from simple grid traversal).

@sahrk has also expressed concerns about the naming scheme and would like a more sane system to standardize on. @isaacbrodsky has suggested this should be part of the version 4.0 update since it would be a backwards-incompatible API change, but I don't see it on the [roadmap](https://github.com/uber/h3/wiki/Roadmap).

I really only have the time to review these code changes, anymore, in between my own work, so take anything I say only as a suggestion, but I think making the API consistent and with a minimal terminology is a good idea and should be on the roadmap. I think perhaps an IETF-like RFC proposal for the change would be worthwhile and a W3C-like requirement of a demo implementation would also be worthwhile.

But I don't think that demo should be done in one of the official bindings like in this repo right now. It will just confuse the Python users when they go to the documentation at uber.github.io/h3 and see a wildly different API reference. I think this library should try to stick as close as possible to the C library and an alternate API wrap around the binding in whichever language the person making the proposal is most comfortable in.

While I'm not recommending this API, [this repo is one such example](https://github.com/sign0/h3-string) of an interesting alternative API for H3. It's a Javascript library but has a Ruby-like ethos of modifying the core types to make things simpler for the end user. The implementation is a very simple wrapper around existing API functions and I think the general approach will make it easier to working test implementations to play with and decide if it seems consistent and obvious, or too simple, or too baroque, etc.

ajfriend:
Agreed that the final version of this branch is not the right place to experiment with a new API, but names are easy to change, and so I figured I'd play around with it while I had the chance ;)

I'd love to converge on some unambiguous terminology. A first pass wouldn't need to cover *everything* (k_ring, hex_ring, ...). It'd be great to just start with the data "types". Maybe something like:

- **index?**: a valid H3 uint64 (of any `mode`: hex/petagon or uni_edge)
- **cell?**: an H3 hexagon or pentagon (`mode` 1)
- **hexagon**: an H3 hexagon (*not* a pentagon)
- **pentagon**: an H3 pentagon

My only worry about `index` is that it doesn't disambiguate `str` and `uint64` representations.

isaacbrodsky:
I agree with @dfellis about naming and what to do with this PR and @sahrk's comment from that issue. The naming is definitely confusing and something that is intended to be fixed in H3 4.0. Currently the next major version does not have a target release date; maybe we should begin a future development branch on uber/h3 as we are starting to accumulate proposed changes to the API and will need to get feedback on them. While it's not currently on the roadmap (oops), uber/h3#104 was created a while ago to hold some of the ideas for 4.0.

An H3 glossary might be a very useful addition to the documentation site. While we can define how we want to use the terms, the way the terms have actually been used in the library has been inconsistent. I can take an attempt to answer the question specifically about "index": an index is the structure that identifies a cell or an edge. It can be encoded in uint64 or hexadecimal string format (previously there was also a C structure with individual fields.) Maybe we'd want to begin a pull request for a glossary document or begin a glossary issue on uber/h3 and list out the terms we'd like to use going forward. We might want to focus on the Python and build changes in this pull request discussion to keep things a little easier to follow.

ajfriend:
Agreed that this PR is getting huge, and we should focus on the Python and build changes. I'll create a separate issue for the glossary.

h3py/api/_api_template.py
ajfriend: It is called `h3_is_valid` in the default "old" API.

h3py/h3core.pyx
ajfriend: Also, the `old` API doesn't expose this term to the users.

h3py/h3core.pyx
ajfriend: fixed.

h3py/h3core.pyx
ajfriend: fixed

h3py/h3core.pyx
ajfriend: fixed

h3py/util.pyx
ajfriend: it does indeed just pass it along to `calloc`. I'll remove that block to avoid confusion.

h3py/h3core.pyx
ajfriend: how about `edge_cells`?

h3py/api/_api_template.py
ajfriend: I've updated the naming to match the current python implementation.

h3py/api/old.py
ajfriend: I've updated the naming to match the current python implementation.

h3py/h3core.pyx
ajfriend: I've updated the naming to match the current python implementation.

h3py/h3core.pyx
ajfriend: I've updated the naming to match the current python implementation.

h3py/h3core.pyx
```diff
@@ -41,6 +40,8 @@ cpdef int base_cell(H3int h) except -1:
 cpdef int resolution(H3int h) except -1:
     """Returns the resolution of an `h3_address`
     0--15
+
+    todo: does this work for edges? or just cells?
```
isaacbrodsky: It should work for edges

h3py/CMakeLists.txt
isaacbrodsky: There's an example of that in the H3 CMakeLsits.txt.

It looks something like:
```
macro(my_macro_name some_arg some_other_arg)
    message("Macro called with ${some_arg} and ${some_other_arg}")
endmacro()
```

You can read more at https://cmake.org/cmake/help/latest/command/macro.html

h3py/h3core.pyx
ajfriend: cool. updated!

h3py/CMakeLists.txt
ajfriend: I macro'd this thing up, and it seems to work... but I have no idea what I'm doing. It'd be great to have someone with more comfort with Cython/CMake to sign off on this eventually. I don't know that we need to do that in this PR, but definitely before we merge the `cython` branch to `master.`

ajfriend:
And now (as far as my tests are concerned) we have Python 2/3 compatibility!

h3py/util.pyx
```diff
@@ -0,0 +1,133 @@
...
+
+from h3py.libh3 cimport H3int, H3str
+cimport h3py.libh3 as h3c
+
+# todo: should we use C API functions instead? (stringToH3 and h3ToString)
+cpdef H3int hex2int(H3str h):
+    return int(h, 16)
+
+cpdef H3str int2hex(H3int x):
+    return hex(x)[2:].strip('L') # todo: this strip is... not ideal
```
isaacbrodsky: please note in the comment why the `strip('L')` is here

ajfriend: good point. added.

PR #61, opened 2019-09-19
Change directory structure, make tests work

ankitmehta21:
* Setup up the following structure:
```
h3-py/
  src/
    h3/
    h3lib/
  tests/
  setup.py
  ...
```

* Reorganize export structure so that library name is h3 instead of h3py


ajfriend:
Awesome! Thanks for this!

Can you make it so that the files are logged in Git as **moved**, instead of deleted and re-created? That would make any changes easy to see in this diff, and preserve the file change history, along with the reasons behind historical changes.

dfellis:
> Awesome! Thanks for this!
> 
> Can you make it so that the files are logged in Git as **moved**, instead of deleted and re-created? That would make any changes easy to see in this diff, and preserve the file change history, along with the reasons behind historical changes.

Git doesn't have that ability. Any `git mv` is actually just a delete and create internally, and the `git diff` tool uses heuristics to figure out if it was moved or not.

In this case, Ankit's changes were significant enough to break the heuristics. Unfortunately you'll have to pull this branch locally and `diff file1 file2` yourself to compare them.

.travis.yml
```diff
@@ -4,6 +4,6 @@ python:
   - "3.6"
 install:
   - pip install -r requirements-dev.txt
-  - fab bootstrap
+  - CFLAGS='-fPIC' fab bootstrap
```
dfellis: Also, this change is why this branch is breaking on Windows. Microsoft's C compiler toolchain uses a completely different set of flags.

You can look at [this file in my h3-node project](https://github.com/dfellis/h3-node/blob/master/.prebuild.sh) to see how to deal with Microsoft's compiler to accomplish what's needed.

ajfriend:
Also, @ankitmehta21, did you base this diff off of the current `uber:cython` branch? The Cython files look like they're from an older version.

I think that may be why I was seeing all the files as deleted/re-created above :)

.travis.yml
ankitmehta21: The travis build is on xenial and passes because of the presence of the fPIC flag, the appveyor build is the windows one I believe

ankitmehta21: The windows build seems to be failing because cmake cannot find a suitable generator program

isaacbrodsky: I think the file @dfellis linked also shows the flags needed to select the right Visual Studio generator on Windows.

ajfriend:
Maybe just a nit, but would it make sense to pick one of `libh3` or `h3lib`, and use that single name throughout the codebase. I think we currently have `src/h3lib/`, but have `src/libh3.pxd`.

I don't have a preference for either, other than that they be consistent, if that's possible.

ankitmehta21:
> Maybe just a nit, but would it make sense to pick one of `libh3` or `h3lib`, and use that single name throughout the codebase. I think we currently have `src/h3lib/`, but have `src/libh3.pxd`.
> 
> I don't have a preference for either, other than that they be consistent, if that's possible.

Resolved, also I played around with the config to get the windows builds to pass. Thanks @dfellis for the pointers!

src/h3/api/_api_template.py
```diff
@@ -1,20 +1,20 @@
-import h3py.h3core as h3core
-import h3py.util as u
+import h3.h3core as h3core
+import h3.h3utils as u
```
ajfriend: why the change from `util` to `h3utils`?

ankitmehta21: `util` seemed like a pretty generic name, I wanted to disambiguate to protect from potential PyPi clashes(eg: [utils package](https://pypi.org/project/utils/), [util package (defunct)](https://pypi.org/project/util/)) while doing `import util` from the h3 folder

ajfriend: fair enough!

ajfriend:
I'm getting an error when cloning and installing on my local machine:

```bash
[ajfriend@ajfriend-C02S63FKG8WM ~/work/temp/h3-py] (cython-ankit)
$ make init
find src -type d -name '*.egg-info' | xargs rm -r
find . -type f -name '*.pyc' | xargs rm -r
find . -type d -name '*.ipynb_checkpoints' | xargs rm -r
git submodule update --init
virtualenv -p python3 env
Running virtualenv with interpreter /usr/local/bin/python3
Using base prefix '/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7'
/usr/local/lib/python2.7/site-packages/virtualenv.py:1047: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
New python executable in /Users/ajfriend/work/temp/h3-py/env/bin/python3.7
Also creating executable in /Users/ajfriend/work/temp/h3-py/env/bin/python
Installing setuptools, pip, wheel...done.
env/bin/pip install -r requirements-dev.txt
Looking in indexes: https://yoober8:****@pypi.uberinternal.com/index, https://pypi.python.org/simple
Collecting cython (from -r requirements-dev.txt (line 2))
  Using cached https://pypi.uberinternal.com/packages/62/13/2ed2e2005afb114ffe08fbd25d890e3795ca86cebf11c97ff8877da58f28/Cython-0.29.13-cp37-cp37m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl
Collecting cmake (from -r requirements-dev.txt (line 3))
  Using cached https://pypi.uberinternal.com/packages/59/20/d1c7e0df29789e7c3aaef17d9952ed477d197cdb4e72d5e23beaf7f3e0c4/cmake-3.15.2.tar.gz
    ERROR: Command errored out with exit status 1:
     command: /Users/ajfriend/work/temp/h3-py/env/bin/python3.7 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/vg/wbk5fpk93z3cnpfn88yzjdbw0000gn/T/pip-install-bt50fay9/cmake/setup.py'"'"'; __file__='"'"'/private/var/folders/vg/wbk5fpk93z3cnpfn88yzjdbw0000gn/T/pip-install-bt50fay9/cmake/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base pip-egg-info
         cwd: /private/var/folders/vg/wbk5fpk93z3cnpfn88yzjdbw0000gn/T/pip-install-bt50fay9/cmake/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/vg/wbk5fpk93z3cnpfn88yzjdbw0000gn/T/pip-install-bt50fay9/cmake/setup.py", line 7, in <module>
        from skbuild import setup
    ModuleNotFoundError: No module named 'skbuild'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
make: *** [init] Error 1
```

ajfriend:
The error goes away if I remove `cmake` from the `requirements-dev.txt`.

But I guess this then requires the user to have previously installed `cmake` via something like `brew install cmake`. Would having it in `requirements-dev.txt` remove the need for this?

ajfriend:
Looks great to me. Thanks so much @ankitmehta21! I'm happy to merge once tests pass.

PR #62, opened 2019-09-24
travis with multiple python versions

ajfriend:


ajfriend:
Next, I'd like to get some coverage test running.

Then, I'd like to see if we can deduplicate the setup, test, and lint logic. Right now, its duplicated over the `makefile`, `.travis.yml`, and `appveyor.yml`. Is there a way to put it all in the `makefile` and have each CI use that?

PR #63, opened 2019-09-25
Basic code coverage

ajfriend:
Currently only covers `.py` files. Ideally, we'd get this to also compute coverage on Cython files.

### DIFF
--- a/src/h3/h3core.pyx
+++ b/src/h3/h3core.pyx
@@ -1,8 +1,13 @@
 cimport h3lib
-from .h3utils cimport check_addr, check_edge, check_res, create_mv, create_ptr
+from h3lib cimport bool, int64_t, H3int
 
-from cpython cimport bool
-from libc.stdint cimport int64_t
+from .h3utils cimport (
+    check_addr,
+    check_edge,
+    check_res,
+    create_mv,
+    create_ptr
+)
 
 from h3.h3utils import H3ValueError
 from h3.geo import (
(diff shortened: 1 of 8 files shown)


src/h3/h3core.pyx
```diff
@@ -1,8 +1,13 @@
 cimport h3lib
-from .h3utils cimport check_addr, check_edge, check_res, create_mv, create_ptr
-
-from cpython cimport bool
-from libc.stdint cimport int64_t
+from h3lib cimport bool, int64_t, H3int
```
ankitmehta21: Should `bool` and `int64_t` still be imported from `cpython` and `libc.stdint`?

ajfriend: These types seemed out of place in `h3core.pyx`, so I wanted to put all the type declarations in one place. `h3lib.pxd` seemed like the best candidate.

ajfriend:
I took a stab at Cython code coverage, but couldn't quite figure out how to get the build working correctly.

I was also looking at this page: https://medium.com/@dfdeshom/better-test-coverage-workflow-for-cython-modules-631615eb197a

I'll put up Cython code coverage as an issue, and we'll see who gets to it first :)

src/h3/h3core.pyx
ankitmehta21: I don't think it should be an issue to import types in `pyx` files, with this it feels like we're doing `stdint.pxd`->`h3lib.pxd`->`h3core.pyx` instead of `stdint.pxd`->`h3core.pyx`

ajfriend: That's exactly what we're doing ;)

I'm thinking of it more like `(whatever random sources)` -> `h3lib.pxd` -> `h3core.pyx`. All types are then defined in `h3lib.pxd`.

I like that better than having types come from three different sources (`cpython`, `stdint`, `h3lib`) -> `h3core.pyx`.

If other `.pyx` files need to use the same types, they can all import from `h3lib`, without having to remember they're hidden away in `cpython` or `stdint`.

Issue #64, opened 2019-09-25
Add Cython code coverage


ajfriend:
We added code coverage for `.py` files in #63, but this doesn't cover Cython `.pxd` or `.pyx` files. We should get Cython code coverage working.

These might be helpful:

- http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html
- https://medium.com/@dfdeshom/better-test-coverage-workflow-for-cython-modules-631615eb197a

ajfriend:
https://github.com/scikit-build/scikit-build/issues/502

PR #65, opened 2019-09-29
Rework some internal names

ajfriend:
- Changed some internal names to remove redundant `h3` prefix: `h3utils` -> `util`. (Having a `util` or `utils` folder shouldn't cause name conflicts, and [seems to be standard in popular python projects.](https://github.com/dask/dask/blob/master/dask/utils.py))
- Changed `h3core` to `memview_int` to match the API naming scheme we currently have.
- Add the `tests` folder to the list of targets to lint (and get them passing)
- Got the `h3.api.array_int` interface working correctly
- Increase coverage on the different interfaces in `h3.api.*`
- Put my email as `author_email` in `setup.py`, as this Cython branch was a rewrite from scratch. @TengHu, does that seem reasonable? Or, would something else be better? Maybe a generic Uber/H3 email address?

This diff should be ready for review.

Next steps after this diff would probably be what's outlined in #64, #66, #67.


src/h3/api/_api_template.py
```diff
@@ -1,5 +1,5 @@
-import h3.h3core as h3core
-import h3.h3utils as u
+import h3.memview_int as mvi
```
nrabinowitz: Nit: I'm not thrilled with this naming chance - it feels to me like it loses some useful abstraction, unless the `memview` name is conventional for direct access to memory.

ajfriend: Memory view is the name of the actual python object we're returning: https://docs.python.org/3/c-api/memoryview.html
It is indeed just a lightweight Python wrapper around memory.

That follows the convention of `set` or `array` as the collection objects used in other APIs.

Memory views don't require any outside dependencies, but most users would expect something like a numpy array, which is why we provide that interface.

PR #71, opened 2019-10-07
Polyfill working on polygons with holes

ajfriend:
Also added some helper functions allowing users to use polyfill without having to make a full GeoJson Polygon dictionary object.


src/h3/api/_api_template.py
```diff
@@ -150,8 +150,18 @@ def uncompact(hexes, res):
 
         return _out_collection(hu)
 
-    def polyfill(geos, res):
-        mv = mvi.polyfill(geos, res)
+    def polyfill_polygon(outer, holes=None, res=9, order='latlng'):
```
isaacbrodsky: I'd suggest against defaulting the resolution. This promotes a specific resolution to be the default one, but it's not clear that we can know that for the broader H3 community.

If an API where the user doesn't always need to specify the resolution is what you want, I'd suggest a class where the user provides the resolution once and then it's reused.

src/h3/geo.pyx
```diff
@@ -35,6 +35,29 @@ cdef (double, double) coord2geo(h3lib.GeoCoord c):
...
+
+    check_res(res)
+
+    c = geo2coord(lat, lng)
+
+    return h3lib.geoToH3(&c, res)
+
+
+cpdef (double, double) h3_to_geo(h3lib.H3int h) except *:
+    """Reverse lookup an h3 address into a geo-coordinate"""
```
isaacbrodsky: "into the centroid geo-coordinate"

src/h3/api/_api_template.py
ajfriend: Yeah, I thought this was a bit awkward too.

This function signature came out of my desire to break up the `holes` from the `outer` LinearRing. I was finding myself constantly making mistakes when the expected input was `[exterior, hole1, hole2, ...]`. For a simple polygon with no holes, it felt awkward to have to write it like `polyfill_polygon([outer])`. 

This all led to making `outer` the first ordered parameter, and adding `holes` as an optional parameter. You can't have non-keyword parameters **after** keyword parameters, so `res` had to have some default value.

I think there are a few different signatures we could use here. I'd love feedback on what you all think would be best.

1. `polyfill_polygon(outer, holes=None, res=9, order='latlng')`
    - current option
2. `polyfill_polygon(outer, holes=None, res=0, order='latlng')`
    - is `res=0` a better default?
3. `polyfill_polygon(outer, holes=None, res=None, order='latlng')`
    - `res` would now be a **required** keyword parameter; we'd raise an error if `res=None`
4. `polyfill_polygon(outer, res, holes=None, order='latlng')`
    - Eh, maybe. is it weird splitting up the `outer` and `holes` parameters?
    - Might be the cleanest option if the most common use case is for Polygons without holes
5. `polyfill_polygon(res, outer, holes=None, order='latlng')`
    - I think this is kind of what you were suggesting @isaacbrodsky. This makes it easy to curry the function, so you could specify `res` just once.
    - It does feel a little unintitive to have `res` first; and it doesn't match our current API.

src/h3/geo.pyx
ajfriend: updated!

src/h3/api/_api_template.py
isaacbrodsky: Yes, `polyfill_polygon(res, outer, holes=None, order='latlng')` is closest to my thinking. The other option I'd suggest is `polyfill_polygon(outer, holes, res, order='latlng')`, making `holes` a required argument. I could also see `polyfill_polygon(polygon, res)` where `polygon` is some more complicated type that encapsulates `outer`, `holes`, and `order`.

ajfriend: I'm leaning towards option 4, as `outer` and `res` are indeed the only **required** parameters. 

Every new user's **first time** using this function will be on simple polygon with no holes. If they had to guess, they'd probably try something like `polyfill_polygon(latlngs, res)`. Option 4 is nice in that this first try (and probably the most common future use-case) *just works*.

A new user's **second time** using the function might involve a hole or two, but at that point, we can expect them to read the docstring ;)

dfellis: if I can jump in with another option:

`polyfill_polygon(coords, res, order='latlng')` with an internal test for the array depth so you can pass in either `outer` or `[outer, hole1, hole2, ...]`

dfellis: Eg:

```py
def polyfill_polygon(coords, res, order='latlng'):
  if isinstance(coords[0][0], (int, float)): coords = [coords]
  ...
```

hhayama: One vote for AJs #4 `polyfill_polygon(outer, res, holes=None, order='latlng')`

nrabinowitz: FWIW, the JS bindings follow @dfellis's suggestion - we allow either a single polygon loop, or a nested polygon loop, and check the input depth to determine which was provided.

ajfriend: @dfellis That's another good option. I think the downsides are:

- it adds hacky type checking code (e.g., do we only check the first element?), and
- it makes it harder to understand what's going on when reading code that's calling the function; i think option 4 [makes it more explicit](https://www.python.org/dev/peps/pep-0020/).

nrabinowitz: Also strong support for `res=None`, rather than providing a default. This is an easy param to forget, and I'd rather see a hard failure if it's left out.

src/h3/api/_api_template.py
```diff
@@ -150,8 +150,18 @@ def uncompact(hexes, res):
...
+        mv = mvi.polyfill_polygon(outer, holes, res, order)
+
+        return _out_collection(mv)
+
+    def polyfill_geojson(geojson, res=9):
```
nrabinowitz: Ditto on `res` default.

src/h3/api/_api_template.py
```diff
@@ -150,8 +150,18 @@ def uncompact(hexes, res):
 
         return _out_collection(hu)
 
-    def polyfill(geos, res):
-        mv = mvi.polyfill(geos, res)
+    def polyfill_polygon(outer, holes=None, res=9, order='latlng'):
+        mv = mvi.polyfill_polygon(outer, holes, res, order)
```
nrabinowitz: Why is this `order` rather than `geojson_conformant`? The latter also implies closed loops.

src/h3/geo.pyx
```diff
@@ -50,58 +73,100 @@ cdef h3lib.Geofence make_geofence(geos):
...
-
-cpdef h3lib.H3int geo_to_h3(double lat, double lng, int res) except 1:
-    cdef:
-        h3lib.GeoCoord c
+        Parameters
+        ----------
+        outer : list or tuple
+            GeoFence
+            A GeoFence is a sequence of >= 4 (lat, lng) pairs where the last
+            element is the same as the first.
```
nrabinowitz: I don't believe the C lib cares whether the loop is closed or not.

src/h3/geo.pyx
```diff
@@ -112,6 +177,56 @@ def polyfill(geos, int res):
...
+    as the GeoJson standard requires them to be in lng/lat order.
+
+    Parameters
+    ----------
+    geojson : dict
+    res : int
+        The resolution of the output hexagons
+    """
+
+    # todo: this one could handle multipolygons...
```
nrabinowitz: Or feature collections, etc. In JavaScript, we moved this functionality to an entirely separate library (https://github.com/uber/geojson2h3).

src/h3/api/_api_template.py
dfellis: > @dfellis That's another good option. I think the downsides are:
> 
>     * it adds hacky type checking code (e.g., do we only check the first element?), and
> 
>     * it makes it harder to understand what's going on when reading code that's calling the function; i think option 4 [makes it more explicit](https://www.python.org/dev/peps/pep-0020/).

On your first point, I don't think checking the first element is any worse than the current behavior of not checking anything at all?

On your second point, I think that's more a style issue than anything. You could consider the first parameter to be the polygon they want to render, where the type of the polygon can basically be `Number[][] | Number[][][]` depending on if there are holes or not, while your approach requires the end user to do their own introspection into the polygon object they're working with and split it into two variables.

src/h3/geo.pyx
```diff
@@ -112,6 +177,56 @@ def polyfill(geos, int res):
...
+
+
+
+
+
+def polyfill(geojson, int res, geo_json_conformant=False):
+    """ Light wrapper around `polyfill_geojson` to provide backward compatibility.
+    """
+
+    gj = geojson.copy()
```
nrabinowitz: This could be expensive for large input polygons, right? Two thoughts:
- Minimally, move the `copy` operation inside `if not geo_json_conformant`, probably with two separate `return` statements
- Ideally, pass the `geojson_conformant` param down to `make_geofence` and swap the input args to `geo2coord` appropriately. This is what we do in JS - you just change the way you pass the memory to C, you never have to duplicate the input.

src/h3/geo.pyx
nrabinowitz: The more I think about it, the more I'd leave this functionality to a separate lib. There's a fair range of stuff you might want to address w/r/t geojson, including `h3_to_multipolygon` output.

src/h3/geo.pyx
ajfriend: Oooh! Good to know. I'll try that out and add something about that to the docstrings.

@nrabinowitz, is that documented in the C lib? I could do a pull request. Would it belong in `h3api.h.in`?

src/h3/geo.pyx
ajfriend: Hmm. So the current `h3-py` `master` branch has the `h3_set_to_multi_polygon()` function. If we want feature parity, the `cython` branch will have to provide that as well. (I was planning on adding that next.)

Do we need this branch to be fully backwards compatible in terms of `h3_set_to_multi_polygon()`?

src/h3/api/_api_template.py
ajfriend: If the C lib doesn't need closed loops, then we don't need to conform to GeoJSON outside of the (lat, lng) vs (lng, lat) order.

It doesn't look like the C lib cares about the right-hand rule either.

Instead of having `geojson_conformant` as an option, I was thinking of having these functions as being as lax as possible, while the `polyfill_geojson()` function would just imply that the input must conform.

src/h3/api/_api_template.py
ajfriend: > On your first point, I don't think checking the first element is any worse than the current behavior of not checking anything at all?

The logic would exist to disambiguate the input format. If we do option 4, then there's no need for that code, because we only allow one format.

> On your second point, I think that's more a style issue than anything. You could consider the first parameter to be the polygon they want to render, where the type of the polygon can basically be `Number[][] | Number[][][]` depending on if there are holes or not, while your approach requires the end user to do their own introspection into the polygon object they're working with and split it into two variables.

True, but I kind of like the idea of expecting the user to know what input format they're using. I also think the function signature of option 4 is more self-documenting than if we allowed a general `coords` input.

I also think option 4 is a more natural format for the data. The `Number[][][]` type is more of a weird artifact of the GeoJSON spec. Option 4 better matches the underlying C struct.

Whatever we do, we'll always need a code path like option 4, even if it is just a helper function that we run after disambiguating the input.

I'll try to rework things in the form of 4. We can revisit before merging (and maybe think of better function names...)

src/h3/api/_api_template.py
ajfriend: fixed!

src/h3/geo.pyx
ajfriend: Good point!

Resolved this so that it no longer does the copy. While I was at it, I decided that having the `order='latlng'` keyword param was kind of awkward, and it should really be a boolean param. I added `lnglat_order=False` instead. Let me know what you all think.

src/h3/geo.pyx
ajfriend: I added a test to confirm that this is true. It takes a polygon with 2 holes and checks every permutation of each LinearRing being clockwise or counterclockwise and being or not being a closed loop.

src/h3/api/_api_template.py
ajfriend: Also, updated the signature to be `polyfill_polygon(outer, res, holes=None, lnglat_order=False)`

src/h3/geo.pyx
```diff
@@ -35,84 +36,200 @@ cdef (double, double) coord2geo(h3lib.GeoCoord c):
...
+
+    for i, (lat, lng) in enumerate(latlng):
         gf.verts[i] = geo2coord(lat, lng)
 
     return gf
 
 
+cdef free_geofence(h3lib.Geofence* gf):
+    if gf.verts:
+        stdlib.free(gf.verts)
```
isaacbrodsky: nit: the C `free` function on `NULL` is a no-op.

src/h3/geo.pyx
nrabinowitz: `h3_set_to_multi_polygon` is just a reasonable binding representation of `h3SetToLinkedGeo`, which is part of the core C API. The pieces that go beyond the C API are:

**GeoJSON to H3**
- Accepting `MultiPolygon` input coordinates
- Accepting GeoJSON input rather than just coords
- Accepting `FeatureCollection` input

**H3 to GeoJSON**
- Outputting GeoJSON rather than just coords
- Outputting alternative transformations, e.g. a `FeatureCollection` of hexagons

Basically anything that knows about *actual* GeoJSON formats, rather than just coordinates, seems out of scope IMO - that's why we put this in a separate lib for JS.

src/h3/geo.pyx
ajfriend: good point. updated.

src/h3/geo.pyx
ajfriend: I do like the idea of keeping this library minimal. I'll leave these in for now so that we match the current `master` branch functionality, but maybe we can revisit which functions we want to include in 4.0.

PR #74, opened 2019-10-19
Add `h3_set_to_multi_polygon`

ajfriend:


src/h3/CMakeLists.txt
```diff
@@ -17,3 +17,4 @@ endmacro()
 add_cython_file(util)
 add_cython_file(geo)
 add_cython_file(memview_int)
+add_cython_file(to_multipoly)
```
ajfriend: @ankitmehta21 Is there a way to just add all the Cython files automatically?

isaacbrodsky: This can be done with `file(GLOB ...)` in CMake but this seems to be recommended against: https://cmake.org/cmake/help/v3.14/command/file.html?highlight=file#filesystem We moved away from using this in uber/h3 for C files.

src/h3/to_multipoly.pyx
```diff
@@ -0,0 +1,88 @@
+cimport h3lib
+from h3lib cimport H3int
+from .util cimport check_addr
+
+
+# need to define `coord2geo` and `mercator` here due to Cython bug:
+# https://github.com/cython/cython/issues/2745
+cdef (double, double) mercator(double lat, double lng):
```
isaacbrodsky: I'd suggest renaming to something like `constrainCoords` - this doesn't seem to have anything to do with the Mercator projection.

src/h3/to_multipoly.pyx
```diff
@@ -0,0 +1,88 @@
...
+
+
+cdef (double, double) coord2geo(h3lib.GeoCoord c):
+    return mercator(
+        h3lib.radsToDegs(c.lat),
+        h3lib.radsToDegs(c.lng)
+    )
+
+# todo: it's driving me crazy that these three functions are all essentially the same linked list walker...
+# grumble: no way to do iterators in with cdef functions!
```
isaacbrodsky: Is there a convenient way to define an iterator for these types so you can write a list comprehension here instead? Or does that not have much benefit?

src/h3/to_multipoly.pyx
```diff
@@ -0,0 +1,88 @@
...
+    return out
+
+# todo: tuples instead of lists?
+def _to_multi_polygon(const H3int[:] hexes):
+    cdef:
+        h3lib.LinkedGeoPolygon polygon
+
+    # todo: should we have a helper that checks a collection of inputs?
+    for h in hexes:
+        check_addr(h) # todo: maybe should name this `check_cell`?
```
isaacbrodsky: I'd move this comment to the `check_addr` function definition

src/h3/to_multipoly.pyx
```diff
@@ -0,0 +1,88 @@
...
+
+    h3lib.h3SetToLinkedGeo(&hexes[0], len(hexes), &polygon)
```
isaacbrodsky: Is `&hexes[0]` different from `hexes`? How so?

src/h3/to_multipoly.pyx
```diff
@@ -0,0 +1,88 @@
...
+
+    out = walk_polys(&polygon)
+
+    # does this thing dealloc the passed-in poly address?
+    h3lib.destroyLinkedPolygon(&polygon)
```
isaacbrodsky: No: https://github.com/uber/h3/blob/master/src/h3lib/lib/linkedGeo.c#L93

(It's created by the caller and thus it is the callers responsibility to deallocate it.)

src/h3/to_multipoly.pyx
```diff
@@ -0,0 +1,88 @@
...
+
+    return out
+
+
+def _json_loop(loop):
```
isaacbrodsky: `_geojson_loop`?

src/h3/to_multipoly.pyx
nrabinowitz: Is this still needed? I'm pretty sure we don't do it in JS any more - we added it into the C lib, so I think all values should be within the expected range.

src/h3/to_multipoly.pyx
```diff
@@ -0,0 +1,88 @@
...
+    loop = [e[::-1] for e in loop]
+    loop += [loop[0]]
+
+    return loop
+
+
+def h3_set_to_multi_polygon(const H3int[:] hexes, geo_json=False):
+    multipoly = _to_multi_polygon(hexes)
+
+    if geo_json:
```
nrabinowitz: Not required, but it would be more efficient to do this within `_to_multi_polygon`, returning either `lat, lng` or `lng, lat` from `coord2geo` and closing the loop while you walk it. Not sure for Python, but in JS the "convert all coords" loop would involve creating new arrays for every coord and garbage-collecting the old ones, at a non-negligible cost for large structures.

I'd consider adding a benchmark and testing this out. It increases the code complexity to pass `geo_json` around as an argument, but it might be worth it if there's a significant perf benefit.

src/h3/CMakeLists.txt
ajfriend: Makes sense. I'll add a comment, just in case the temptation ever arises again.

src/h3/to_multipoly.pyx
nrabinowitz: Correct, C output is still radians.

src/h3/to_multipoly.pyx
ajfriend: It's a weird Cython-ism. `hexes` isn't a pointer, its a Cython `memoryview` object.

`&hexes[0]` is the standard idiom for getting the pointer to the start of the array.

http://docs.cython.org/en/latest/src/userguide/memoryviews.html#pass-data-from-a-c-function-via-pointer

src/h3/to_multipoly.pyx
ajfriend: Cool.

I guess I was confused by the pass-by-reference. Couldn't we just pass this by value?

src/h3/to_multipoly.pyx
ajfriend: done

src/h3/to_multipoly.pyx
ajfriend: got it!

src/h3/to_multipoly.pyx
ajfriend: just went ahead and renamed the function

src/h3/to_multipoly.pyx
ajfriend: I'm on the same page. I wasn't sure about the performance gain, so I leaned towards the code complexity savings.

How about I make an issue for us to benchmark this?

A possible approach to keep the complexity down while not sacrificing speed would be to have `_to_multi_polygon` return iterators of iterators (of iterators...?). That would let us defer all the loop-closing, lat/lng swapping, and `tuple` vs. `list` logic.

ajfriend: Created: https://github.com/uber/h3-py/issues/76

src/h3/to_multipoly.pyx
ajfriend: That's what I was really hoping to do! :)

The functions would be so clean:

```python
cdef walk_polys(const h3lib.LinkedGeoPolygon* L):
    while L:
        yield walk_loops(L.data)
        L = L.next
```

Unfortunately, a `cdef` function is essentially a C function, and [so it can't be a generator function (with a `yield` statement)](https://github.com/cython/cython/issues/2597).

We could turn it into a normal Python function to use the `yield` statement, and [Python functions **can** take in C pointers, but only if Cython can figure out how to do the conversion automatically](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#python-functions-vs-c-functions). It seems these objects are too complicated. It might still be possible, I just couldn't figure out how to do it.

The other option would be to crate a Python/Cython object to wrap the `h3lib.LinkedGeoPolygon*`, but at that point, it seemed to be creating more complexity than it would be saving...

src/h3/to_multipoly.pyx
isaacbrodsky: It could be pass by value, but generally we pass structures by reference in C (or by constant reference) to eliminate the need to push the entire structure onto the stack.

ajfriend: got it!

src/h3/memview_int.pyx
```diff
@@ -74,8 +76,8 @@ cpdef H3int parent(H3int h, res=None) except 1:
 cpdef int distance(H3int h1, H3int h2) except -1:
     """ compute the hex-distance between two hexagons
     """
-    check_addr(h1)
-    check_addr(h2)
+    check_cell(h1)
+    check_cell(h2)
```
isaacbrodsky: nit: will this produce an error that mentions the argument name? Might want to mention either the first or second index are invalid

src/h3/to_multipoly.pyx
```diff
@@ -0,0 +1,80 @@
+cimport h3lib
+from h3lib cimport H3int
+from .util cimport check_cell
+
+
+# need to define `coord2geo` here due to Cython bug:
+# https://github.com/cython/cython/issues/2745
+cdef (double, double) coord2geo(h3lib.GeoCoord c):
```
isaacbrodsky: `coord2degs`?

src/h3/to_multipoly.pyx
isaacbrodsky: Interesting, thanks for sharing!

src/h3/memview_int.pyx
ajfriend: That's a good point. It currently doesn't. Even worse, the error is reported in decimal instead of the hexadecimal string.

I don't see a quick fix off the top of my head, so I'll move this to an issue to think about more deeply.

ajfriend: Added: https://github.com/uber/h3-py/issues/78

src/h3/to_multipoly.pyx
ajfriend: I tried something with [Cython fused types](https://cython.readthedocs.io/en/latest/src/userguide/fusedtypes.html), but couldn't get it working.

Will revisit later.

src/h3/to_multipoly.pyx
ajfriend: I `deg` it.  :sunglasses:

Updated, but without the 's'.

PR #77, opened 2019-10-21
README change to mention Git and Cmake version

### LINKED ISSUE #75: Clarify that installation expects git and cmake (version 3)
Hi, it would be nice to clarify in the README that git and cmake (version 3) are needed for installation. I came across installation errors (triggered by the H3 install bash [script](https://github.com/uber/h3-py/blob/392fe162f47e4d2ab33f0edde4786fe017bbd966/.install.sh)) when pip-installing h3 in a Linux-based production environment:
* git not available (though we might reasonably assume git is available in a local dev environment)
* cmake installed by `sudo yum install cmake` was version 2, but apparently H3 needs version 3?

My issues were resolved by `sudo yum install -y cmake3 git && sudo ln -s /usr/bin/cmake3 /usr/bin/cmake`. I realize that whether one encounters these issues, as well as how they can be resolved, depends on the use case, but the need for git and cmake (version 3) seems worth calling out in the documentation.

Thanks for developing H3 and the python wrapper!

isaacbrodsky:
Fixes #75. Also added a small note about Windows installation (from issues this might need more explanation as to whether it can be installed except through Git Bash).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,8 +10,8 @@ the `H3 Documentation <https://uber.github.io/h3/>`__.
 Installing
 ==========
 
-You need to have ``cc``, ``make``, and ``cmake`` in your ``$PATH`` when installing this
-package:
+You need to have ``cc``, ``make``, ``cmake`` (version 3.1 or above),
+and ``git`` in your ``$PATH`` when installing this package:
 
 .. code:: sh
 
@@ -21,10 +21,18 @@ package:
     /usr/bin/make
     which cmake
     /usr/bin/cmake
+    which git
+    /usr/bin/git
 
 .. code:: sh
 
     pip install h3
+
+Installing on Windows
+---------------------
+
+On Windows you will also need to have ``bash`` (for example from Git)
+installed.
 
 Development
 ===========
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,7 @@
 # Testing
-pytest==4.0.2
-pytest-cov==1.8.1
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -21,11 +21,19 @@ package:
...
+    /usr/bin/git
 
 .. code:: sh
 
     pip install h3
 
+Installing on Windows
+---------------------
+
+On Windows you will also need to have ``bash`` (for example from Git)
```
nrabinowitz: `bash` specifically? Or some shell?

ajfriend:
LGTM! Ideally, with the Cython rewrite we'll publish wheels, which should remove the need for `make`, `git`, `cmake`, and a compiler (as long as people install from PyPI).

README.rst
isaacbrodsky: `setup.py` specifically tries to invoke `bash`, it's possible another *nix shell will work, but it will need to alias itself to `bash`.

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 # Testing
-pytest==4.0.2
-pytest-cov==1.8.1
-coverage==3.7.1
+pytest==4.6.6
+pytest-cov==2.6.1
+coverage==4.5.4
```
isaacbrodsky: Upgrading the test dependencies is needed because a transitive dependency broke (same error as https://stackoverflow.com/questions/58189683/typeerror-attrib-got-an-unexpected-keyword-argument-convert)

PR #79, opened 2019-10-22
Get original test suite working

ajfriend:


src/h3/memview_int.pyx
```diff
@@ -147,6 +148,10 @@ cpdef H3int[:] children(H3int h, res=None):
 
 
 cpdef H3int[:] compact(const H3int[:] hu):
+    # todo: gotta be a more elegant way to handle these...
+    if len(hu) == 0:
+        return empty_memory_view()
```
isaacbrodsky: What goes wrong if `hu` is zero length? If it's a core library crash, that should be addressed by the yet unreleased uber/h3#278

src/h3/memview_int.pyx
```diff
@@ -162,6 +167,10 @@ cpdef H3int[:] compact(const H3int[:] hu):
 # todo: https://stackoverflow.com/questions/50684977/cython-exception-type-for-a-function-returning-a-typed-memoryview
 # apparently, memoryviews are python objects, so we don't need to do the except clause
 cpdef H3int[:] uncompact(const H3int[:] hc, int res):
+    # todo: gotta be a more elegant way to handle these...
+    if len(hc) == 0:
+        return empty_memory_view()
```
isaacbrodsky: Same question as above

tests/test_h3.py
```diff
@@ -0,0 +1,1072 @@
...
+
+    # note: changed from master
+    # note: not sure what the correct return value is here...
+    def test_silly_geo_to_h3(self):
+        out = h3.geo_to_h3(37.3615593 + 180.0, -122.0553238 + 360.0, 5)
+
+        expected_master_branch = '85283473fffffff'
+        expected_cython_branch = '85ca2d53fffffff'
+
+        assert out == expected_master_branch
```
ajfriend: This is the last test that's failing for me.

The current `master` branch expects `'85283473fffffff'`.
The new `cython` branch returns `'85ca2d53fffffff'`.

Any idea which one is the correct result? @isaacbrodsky @nrabinowitz @dfellis?

ajfriend: Could this have anything to do with me dropping the ["mercator" transform in the last commit](https://github.com/uber/h3-py/commit/963241447bba1151958199c6ac30d726c34a2345#diff-1df158fef384a0b5e893d1ef4a14c773L13-L20)?

We did that because we decided that should have been handled by the C library. We sure that's the case?

tests/test.py
```diff
@@ -381,11 +381,11 @@ def test_line():
 
     out = h3.h3_line(h1, h2)
 
-    expected = {
+    expected = [
         '8928308280fffff',
         '89283082873ffff',
         '8928308287bffff'
-    }
+    ]
```
ajfriend: @nrabinowitz Should the output of `h3_line` be considered an **ordered** list or an **unordered** set?

ajfriend:
Also, in porting over the tests from `master`, some of those tests no longer needed to fail, so I changed those tests.

Any tests I changed, I tried to mark with `# note: changed from master`. Since I just copied the test file over from `master`, I couldn't see an easy way to make those diffs obvious other than commenting.

tests/test_h3.py
isaacbrodsky: The first index (master) is what the C library gives, too. I also tested an older version of the core library and got the same result. However, the "mercator" function seems to be wrapping the north pole (90 degrees) around to the south pole (-90 degrees), which I'm not sure makes sense. I think this behavior may be correct.

tests/test.py
isaacbrodsky: The guarantees for this function are at https://github.com/uber/h3/blob/master/src/h3lib/lib/localij.c#L591 The output is ordered because "... every index in the line will be a neighbor of the preceding index."

tests/test_h3.py
nrabinowitz: The JS lib gives `85ca2d53fffffff` (Cython branch). Not sure why this would be different from the C library :/.

ajfriend: It's a lat/lng normalization issue. I'm pretty sure `85283473fffffff` is the correct cell here.

I don't know why it was called that, but the `mercator()` function that we previously removed just tried to ensure that `lat` is in the interval `[-90, 90]` (should it be half-open?) and `lng` is in `[-180, 180]`.

The function would thus transform:

```
(37.3615593 + 180.0, -122.0553238 + 360.0) -> (37.3615593, -122.0553238)
```

Weird that the C library seems to do this normalization correctly, but the Cython and JS libs do not. Especially since they're just calling the C lib.

@isaacbrodsky, how did you test this with the C lib? Any chance there's some pre-processing code that is normalizing the coordinates?

tests/test.py
ajfriend: Got it. So there's no need to check for and remove zeros in the output?

tests/test_h3.py
ajfriend: Maybe we want a function like this?

```python
def normalize(lat, lng):
    """ Helper to coerce lat/lng range
    
    lat in [-90, +90]
    lng in [-180, +180]
    
    """
    while lat > +90:
        lat -= 180
    
    while lat < -90:
        lat += 180

    while lng > +180:
        lng -= 360
        
    while lng < -180:
        lng += 360

    return lat, lng
```

We could also just do some modulus operation, but I think this makes the logic clear.

ajfriend: Alternatively, I guess we could also just throw an error if the inputs are outside of expected bounds?

isaacbrodsky: I tested using the command line `geoToH3` program. It just converts the coordinates to radians and feeds them into heoToH3. I don't think that wrapping to the south pole makes sense.

dfellis: Right, but wrapping back the other direction also doesn't make sense without changing the longitude at the same time. If we treat the coordinates as a vector from `(0, 0)` then getting to a latitude of 91 degrees really implies a latitude of 89 degrees while also adding 180 degrees to the longitude and clamping again, right?

ajfriend: I'll resolve this for now by adding back in the `mercator()` function, so that the Cython branch agrees with `master` and the C library.

Kicking the discussion of what we should **really** do to https://github.com/uber/h3/issues/287.

tests/test.py
isaacbrodsky: I don't think the output from the `h3Line` function can contain zeros.

ajfriend:
All the original tests from master should now be passing. (Some tests were modified because failures were no longer needed.)

Next up: lint the `test_h3.py` file so that everything can pass once and for all!

tests/test_h3.py
ajfriend: Actually, I've switched my opinion on this. Notes here: https://github.com/uber/h3/issues/287

tests/test.py
ajfriend: Thinking about this now, I wonder if `h3Path` would be a better name, to indicate to the user that this returns a path between two nodes in the cell adjacency graph.

Maybe something to consider when we look at naming in 4.0?

ajfriend: Resolved this with https://github.com/uber/h3-py/pull/79/commits/32fe6a0ff34a1513f8fb038b2b1fb1f4d4ded908

src/h3/memview_int.pyx
ajfriend: The bug appears when I run the test suite (without the above `if` block) as:

```python
E   IndexError: Out of bounds on buffer access (axis 0)
memview_int.pyx:207: IndexError
```

It looks like it comes down to this expression: `&hc[0]`. This is the standard syntax for Cython, so that's tricky to avoid.

It sounds like the change in uber/h3#278 was to not dereference the input pointer if the length was zero. Is that right?

If that's the case, given that the Cython syntax essentially **requires** dereferencing, I think the current solution which short-circuits the whole mess is probably the easiest.

Anyone see a cleaner solution?

@isaacbrodsky @nrabinowitz @dfellis @karentycoon

src/h3/api/_api_template.py
```diff
@@ -114,12 +115,38 @@ def h3_to_geo_boundary(h, geo_json=False):
...
+    def hex_ranges(hexes, K):
+        out = {
+            h: hex_range_distances(h, K)
+            for h in hexes
+        }
+
+        return out
+
+    def k_ring_distances(h, K):
+        return hex_range_distances(h, K)
```
isaacbrodsky: Not sure about a `k_ring` function that calls hex_range, seems like a naming issue.

src/h3/memview_int.pyx
isaacbrodsky: Maybe add a brief comment explaining this is good. This solution seems good.

src/h3/api/_api_template.py
ajfriend: Agreed. I've never gotten the distinction between `k_*` and `hex_*`,  especially with `*_ring` and `*_range`.  Do these things correspond to implementation details (around hitting pentagons)?

Seeing as most users don't even know that pentagons are a possibility, would it make sense to focus the "external" API towards "cell graph adjacency" and the "internal" functions towards the implementation details of pentagons and icosahedron faces?

ajfriend: Added a comment to the code on this solution/choice here: https://github.com/uber/h3-py/pull/79/commits/05dd65178f3ced0e98ceb61fe7cd2cc765d9b22f

src/h3/memview_int.pyx
ajfriend: Added a comment here: https://github.com/uber/h3-py/pull/79/commits/05dd65178f3ced0e98ceb61fe7cd2cc765d9b22f

PR #83, opened 2019-11-28
Get tox working

ajfriend:
- `tox` makes it easy to test multiple environments automatically
- I had to remove `pyproject.toml` to get it working. What was it for? It was messing up install paths like `pip install .`. I'm putting up this PR partially to see if it messes up our CI builds
- I simplified the dynamic module loading a bit
- `tox` only works locally for now, I still have to migrate the CIs to use `tox`

The problems I was getting with `pyproject.toml` are probably explained in this post: https://snarky.ca/clarifying-pep-518/


pyproject.toml
```diff
@@ -1,2 +0,0 @@
-[build-system]
-requires = ["setuptools", "wheel", "scikit-build", "cmake==3.14.4", "ninja"]
```
ajfriend: @ankitmehta21 What did we need this for?

ajfriend:
I think this PR is basically where I'd like it to be.

Next steps:

- move all the `.pyx` files to a `_cy` folder (or a better name) to isolate the Cython code and to better hide those functions from the user (when tab completing, for example) @ankitmehta21, I might need your help here, as I took a stab, but couldn't get the CMake/build stuff to work.
- I'd like to move `test_h3.py` to the lighter-weight `pytest` syntax.

PR #84, opened 2019-12-03
Convert original test suite to use pytest syntax

ajfriend:
Test semantics should all be the same. The only change here is to move from `unittest` syntax to `pytest` syntax, which is a bit lighter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,6 @@ install:
   - CFLAGS='-fPIC' python setup.py install
 script:
   - flake8 src/h3 setup.py tests
-  - pytest tests/* --cov=h3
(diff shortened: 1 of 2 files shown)


nrabinowitz:
I have the same concern @dfellis expressed in Slack - any way to keep the old messages, e.g. as comments? Otherwise you lose some of the documentation power of the tests.

ajfriend:
@dfellis @nrabinowitz, the old messages should all (except for the ones that I thought were obvious) be preserved in the form of comments, and thus visible whenever we hit an error in the test suite.

Let me know if that looks OK, or if I should add any other comments!

PR #86, opened 2019-12-07
Slightly improve invalid index error reporting

ajfriend:
- it is now fairly easy to `pip install` from the github repo to test things out!
- reworked imports so that they're now relative
- start to do a better job of hiding internal names from the end user

## next step

- I'd like to hide the internal Cython modules from the user by putting them into something like a `_cy` folder. However, I can't quite get that working with Cython/CMake/scikit-build, so I'll put up a (failing) PR with the basic structure I'd like and see if I can get some help to get it working.


readme.md
```diff
@@ -21,6 +21,11 @@ Try from within `ipython`:
 '8075fffffffffff'
 ```
 
+## Install from github
```
isaacbrodsky: nit: `GitHub` (capitalize)

src/h3/_internal_api.py
```diff
@@ -2,7 +2,7 @@
 
 """
 This module should serve as the interface between the C/Cython code and
-the Python code. That is, this module should import all the Cython functions we
+the Python code, i.e., an internal API. That is, this module should import all the Cython functions we
```
isaacbrodsky: nit suggestion: "... the Python code, that is, it is an internal API. This module should import all the Cython functions we ..."

src/h3/api/_api_template.py
```diff
@@ -1,4 +1,4 @@
-import h3.core as c
+from .. import _internal_api as _a
```
isaacbrodsky: I am a little concerned about names like `c` or `_a`. They are not descriptive and I think this name has already been renamed. Is there an abstract name for what this is?

src/h3/api/_api_template.py
```diff
@@ -17,22 +17,42 @@ def _api_functions(
         _globals,
 ):
     def versions():
-        return c.versions()
+        """ Return version numbers for both the Python library
+        and the underlying H3 C library.
+
+        Returns
+        -------
+        dict
```
isaacbrodsky: Document what are the keys?

src/h3/api/_api_template.py
```diff
@@ -17,22 +17,42 @@ def _api_functions(
...
+        """ Converts an H3 64-bit integer index to a hexadecimal string
+        """
+        return _a.int2hex(x)
 
     def num_hexagons(resolution):
-        return c.num_hexagons(resolution)
+        """ Return the total number of *cells* (hexagons and pentagons)
+        for the given resolution.
+
+        todo: is that right?
```
isaacbrodsky: Yes, you can confirm this by checking that the number at res 0 is 122.

src/h3/util.pyx
```diff
@@ -25,7 +25,7 @@ def versions():
     return v
 
 
-cpdef H3int hex2int(H3str h):
+cpdef H3int hex2int(H3str h) except *:
```
isaacbrodsky: Per https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html could `except? 0` be used here?

src/h3/api/_api_template.py
ajfriend: I have a whole philosophy around this ;)

I think longer, descriptive names are important when you have a large namespace to hold in your head and the names help you remember what that object is. An alternative strategy is to keep namespaces small so you don't have to keep track of lots of things.

In this file there's only one "thing" being imported from the outside, so I'd say its pretty easy to remember what's going on. Pretty much any name would work, but I think shorter names keep the code easier to read. (I'm even a little upset I have to use two whole characters here, but the underscore helps hide the symbol from the end user.)

The example I always think of is a squaring function like

```python
def square(thing_to_be_squared):
    result_of_squaring = thing_to_be_squared*thing_to_be_squared

    return result_of_squaring
```

when something like

```python
def square(x):
    return x*x
```

is probably **more** clear, even though `x` isn't as descriptive of a name. The namespace of the function is so small that the name doesn't really matter, as everything can be figured out from context.

That said, I'm still completely open if you have other suggestions!

src/h3/api/_api_template.py
ajfriend: good point. will add!

src/h3/api/_api_template.py
ajfriend: cool!

src/h3/util.pyx
ajfriend: good point! I'll change that

readme.md
ajfriend: changed!

src/h3/_internal_api.py
ajfriend: updated

src/h3/api/_api_template.py
isaacbrodsky: I understand that, although I would prefer we avoid renaming this in future since it creates large diffs.

src/h3/util.pyx
isaacbrodsky: Might also apply in other places like `src/h3/util.pyx ` line 75, but that's not new in this PR.

src/h3/api/_api_template.py
ajfriend: Good point. I'm still "getting a feel" for the right way to design this wrapper, so a few large diffs might still happen as I rearrange things. I've been considering that the cutoff point for that kind of refactoring would be when we merge this branch into `master`.

src/h3/util.pyx
ajfriend: Oooh, I totally didn't realize we had the `except?` syntax!

I'm now using `cpdef H3int hex2int(H3str h) except 0`, which should work, assuming that no valid H3 index can ever be zero.

In `util.py`, I'm thinking of using `cdef H3int* create_ptr(size_t n) except? NULL`, since `NULL` could potentially be a valid output if `n = 0`.

For the two examples...
- `edges.pyx`, `cpdef (H3int, H3int) edge_cells(H3int e) except *:`
- `geo.pyx`, `cpdef (double, double) h3_to_geo(H3int h) except *:`

...I think we have to stick with the `except *`, as I can't figure out how to get the syntax to work with anything else for multiple return values.

I would think that multiple return values would evaluate to a tuple (which is a Python object) making the `except *` unnecessary, but that doesn't see to be the case (the tests expecting the error to appear fail). And something like `except (0,0)` doesn't seem to work either. Happy to change it if anyone knows a better way!

src/h3/api/_api_template.py
```diff
@@ -17,22 +17,43 @@ def _api_functions(
         _globals,
 ):
     def versions():
-        return c.versions()
+        """ Return version numbers for both the Python library
+        and the underlying H3 C library.
+
+        The two versions should match exactly.
```
nrabinowitz: Is this correct? Our versioning strategy has been to allow the bindings to set their own patch versions - otherwise it's difficult to bump the binding for binding-specific fixes.

src/h3/api/_api_template.py
```diff
@@ -17,22 +17,43 @@ def _api_functions(
...
 
     def h3_to_string(x):
-        return c.int2hex(x)
+        """ Converts an H3 64-bit integer index to a hexadecimal string
+        """
+        return _a.int2hex(x)
 
     def num_hexagons(resolution):
-        return c.num_hexagons(resolution)
+        """ Return the total number of *cells* (hexagons and pentagons)
```
nrabinowitz: That's a good point - this should be `numIndexes` in the core lib in v4.

src/h3/api/_api_template.py
dfellis: Yep, I agree with @nrabinowitz the major and minor should match, but patch is freely updated on the bindings.

ajfriend: Got it. I'll update.

src/h3/api/_api_template.py
ajfriend: What about `numCells`? An "index" could still be a unidirectional edge, right?

nrabinowitz: I think there's some general terminology we ought to work out for v4... currently we've been using `index` consistently for cell index, but it might be worth refining.

PR #88, opened 2019-12-10
Remove version checks in CMakeLists.txt

ajfriend:
Since we're using `git submodule`, we don't need the version check code in `CMakeLists.txt`.

One potential question, however, is if we should be using `git submodule`, or an alternative like `git subtree`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,18 +1,6 @@
 cmake_minimum_required(VERSION 3.7.2)
 
-set(version_regex "([0-9.]+)")
-file(STRINGS "src/h3lib/VERSION" h3_version
-     REGEX "^${version_regex}$")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,18 +1,6 @@
 cmake_minimum_required(VERSION 3.7.2)
 
-set(version_regex "([0-9.]+)")
-file(STRINGS "src/h3lib/VERSION" h3_version
```
dfellis: You should also see if you can remove the `h3_version` file from the repo successfully.

ajfriend: There isn't one. I'm guessing that's an older file, and we never updated this line?

dfellis: oh, sorry, I misread this line. nevermind!

PR #89, opened 2019-12-10
Better API names

ajfriend:
Change API names:

- `set_int` -> `py_int`
- `set_str` -> `py_str`
- `array_int` -> `numpy_int`

How do people like those names? I'm not sure I love the `py_` prefix. I'd like to just convey that these modules use standard Python data types like `set` and `tuple`, instead of the more exotic types like `numpy.array` and `memoryview`.

Would something like `basic_int`, `standard_int`, `std_int` be better?


ajfriend:
Hmmm... thinking through the API definitions has surfaced this issue: #90

src/h3/api/basic_str.py
```diff
@@ -16,10 +25,12 @@ def _in_collection(hexes):
 
 
 def _out_unordered(mv):
+    # todo: should this be an (immutable) frozenset?
```
isaacbrodsky: Is there a preference in Python for working with the mutable types? Otherwise my preference would be the immutable types.

ajfriend: I'd prefer immutable types. In Python `set` is mutable, while [`frozenset`](https://docs.python.org/3/library/stdtypes.html?#frozenset) is immutable.

The only things holding me back from using `frozenset` are that:

- `frozenset` is a bit more exotic than `set`, so fewer users will be familiar with it
- `set` has a less-verbose syntax and representation than `frozenset`: `{1,2,3}` vs `frozenset({1, 2, 3})`, so we'll be seeing a lot of `frozenset({...})` typed out in documentation, examples, and tests

However, for the last point, `frozenset([1,2,3]) == {1,2,3}`, so we could use the lighter `set` syntax for tests, at least.

Also, if we do this for **unordered** output, we should probably make the **ordered** output immutable. We'd just need to make sure we're outputting `tuple` instead of `list` everywhere else.

ajfriend: Ooh! I should also look into if we can set the `memoryview` and `np.array` outputs as immutable.

dfellis: I'm not as convinced about forcing immutable inputs and outputs. Why should we stop someone who wants to update the output they received from this library? At the very least I would want an option to disable this behavior.

ajfriend: Also, `[1,2,3] != (1,2,3)`, so switching to`tuples` for ordered output probably breaks the API.

ajfriend: @dfellis, I do like the benefits of immutable objects in terms of reasoning about code.

If you do want to modify the object, a user can always make a mutable copy. I can't immediately think of any use cases where the copy would be performance limiting. (An explicit copy also makes the intention of the code more clear!)

In terms of giving the user the option, I suppose we could provide mutable and immutable versions of the APIs.

dfellis: So making a copy of data always has a performance penalty versus not, as the computer has to allocate new memory (which sometimes has to call out to the kernel and punt your application to the back of the kernel scheduler's line) and then copy element-by-element over. That's why I suggested having an option to turn immutability off.

I understand the value of immutability for reasoning about code, but when the language doesn't understand that concept at a fundamental level where it can bake in optimizations (like under-the-hood mutating a data structure because it knows the original is never referenced again), it causes a lot of memory thrashing.

I normally wouldn't bring this up for the Python bindings since I don't believe Python is the right language to use if you care for performance, but since this Cython rewrite with baked-in support for Numpy and raw memory blocks is very much about improving the performance of the Python bindings, I think it's worth keeping the last potential bottleneck on performance as an optional behavior, especially since immutability wasn't enforced before.

ajfriend: Sure, I should have said that I don't think copying would usually be the **bottleneck** in Python. And I don't think the small cost of copy outweighs the clarity benefits of immutability. Basically, I think it's a premature optimization. 

The numpy binding does get a speed benefit from the fact that we don't copy the memory from C, and instead re-uses it in the numpy array or memory view.

In the numpy case, however, I can't imagine a use case where someone would want to work on an array of H3 indexes **in place**. That sort of thing makes sense in something like a linear algebra library, but I'm not sure I see it for a fixed-length set of hexagons.

I'm open to adding it as an option, but I would love to see a use case for it first. 

That said, for this first release, I think we need to keep `set` and `list` instead of their immutable counterparts, just so we match the current API.

dfellis: I can think of an immediate use-case, though. The indexes do represent a quasi-2D space, and I could see someone applying a vector offset transform on an array of hexes to get the same set shifted by some distance, or an array of vectors applied to the hexes for forecasting future concentrations based on current transition vectors, inline replacing the hex IDs.

Basically any sort of `map`-like operation where the original IDs are not needed afterwards could benefit from a mutable list or set.

ajfriend: Fair enough. And again, I think we're stuck with `list` and `set` for the immediate future, but can think about this issue in terms of the 4.0 release.

For 4.0, how about this:

We make the standard interfaces `basic_int`, `basic_str`, `array_int` output immutable collections. 

The remaining interface, `memview_int`, is intended to be as low-level as is reasonably possible. Thinking of this as the "advanced" interface, we can leave the output as mutable arrays of `uint64`s, and let the user do what they'd like.

This way, @dfellis, the default interfaces get the benefits of immutability, and if you want to do something that requires mutability, you're free to implement it using the bare-bones `memview_int` interface.

Since that's a 4.0 issue, I think we're ready to land this diff for now, unless anyone has objections.

PR #91, opened 2019-12-29
Add more docstrings to the APIs

ajfriend:


src/h3/api/_api_template.py
```diff
@@ -10,6 +10,12 @@
 from .. import _cy
 
 
+# H3 address vs H3 cell vs H3 Idnex?
+# do we want a better type system here?
+# how to describe `Set[H3 Index]`, have type distinguish between
+# cells and edges?
```
isaacbrodsky: typo: idnex -> index

I'd suggest moving this to an issue here or on uber/h3.

src/h3/api/_api_template.py
```diff
@@ -61,9 +67,21 @@ def num_hexagons(resolution):
         return _cy.num_hexagons(resolution)
 
     def hex_area(resolution, unit='km^2'):
+        """ Return the average area of an H3 cell
+        for the given resolution.
+
+        todo: all cells, or just hexagons?
```
isaacbrodsky: This is for hexagonal cells only (see also https://github.com/uber/h3/issues/168)

src/h3/api/_api_template.py
```diff
@@ -61,9 +67,21 @@ def num_hexagons(resolution):
...
+        todo: `mean_hex_area`?
+        """
         return _cy.mean_hex_area(resolution, unit)
 
     def edge_length(resolution, unit='km'):
+        """ Return the average cell edge length
+        for the given resolution.
+
+        todo: all cells, or just hexagons?
```
isaacbrodsky: Presumably also for only hexagonal cells

src/h3/api/_api_template.py
```diff
@@ -120,10 +160,12 @@ def h3_to_parent(h, res=None):
         return p
 
     def h3_distance(h1, h2):
-        """ Compute the hex-distance between two hexagons
+        """ Compute the H3-distance between two cells
```
isaacbrodsky: "H3-distance" -> "distance in number of grid cells"?

src/h3/api/_api_template.py
```diff
@@ -136,24 +178,65 @@ def h3_distance(h1, h2):
         return d
 
     def h3_to_geo_boundary(h, geo_json=False):
+        """ Return tuple of lat/lng paris describing
```
isaacbrodsky: typo: paris -> pairs

src/h3/api/_api_template.py
```diff
@@ -136,24 +178,65 @@ def h3_distance(h1, h2):
...
+        the cell boundary
+
+        If `geo_json` is True, return lng/lat pairs
```
isaacbrodsky: We should remark on what the ordering is if this is false.

src/h3/api/_api_template.py
```diff
@@ -136,24 +178,65 @@ def h3_distance(h1, h2):
...
+        Returns
+        -------
+        Set[H3 Index]
+        """
         mv = _cy.disk(_in_scalar(h), k)
 
         return _out_unordered(mv)
 
     def hex_range(h, k=1):
+        """ Alias for `k_ring`. "Full" disk.
```
isaacbrodsky: Note that this differs from the C API - hexRange is a different function there.

src/h3/api/_api_template.py
```diff
@@ -172,32 +267,50 @@ def hex_ranges(hexes, K):
...
-        collection of h3 addresses
+        UnorderedCollection[H3Cells]
         """
         mv = _cy.children(_in_scalar(h), res)
 
         return _out_unordered(mv)
 
     # todo: nogil for expensive C operation?
     def compact(hexes):
+        """ Compact a collection of H3 Cells by combining
```
isaacbrodsky: nit: Cells -> cells? I'm a little unclear if it's correct to capitalize here as you're using this as a type name.

src/h3/api/_api_template.py
```diff
@@ -172,32 +267,50 @@ def hex_ranges(hexes, K):
...
         hu = _in_collection(hexes)
         hc = _cy.compact(hu)
 
         return _out_unordered(hc)
 
     def uncompact(hexes, res):
+        """ Reverse the `compact` operation and return a collection
+        of H3 Cells, all of resolution `res`.
+
+        what if uncompact input contains a hex samller than res?
```
isaacbrodsky: An error code will be returned: https://github.com/uber/h3/blob/master/src/h3lib/lib/h3Index.c#L425

src/h3/api/_api_template.py
```diff
@@ -223,28 +336,53 @@ def polyfill(geojson, res, geo_json_conformant=False):
         return _out_unordered(mv)
 
     def h3_is_pentagon(h):
-        """
-        a pentagon should still pass is_cell(), right?
-        :returns: boolean
+        """ Returns `True` if input is a valid H3 Cell which is
+        a pentagon.
+
+        todo: a pentagon should still pass is_cell(), right?
```
isaacbrodsky: Yes, a pentagon is a valid cell.

src/h3/api/_api_template.py
```diff
@@ -223,28 +336,53 @@ def polyfill(geojson, res, geo_json_conformant=False):
...
     def h3_get_base_cell(h):
-        """
-        :returns: boolean
+        """ Return the parenting cell of resolution `0`.
+
+        todo/question: overlap with the `get_parent()` function?
+
+        Returns
+        -------
+        bool
```
isaacbrodsky: This returns an integer base cell number, not a bool (and also not a parent index - it just reads a bitfield from the index.)

src/h3/api/_api_template.py
```diff
@@ -223,28 +336,53 @@ def polyfill(geojson, res, geo_json_conformant=False):
...
         return _cy.get_base_cell(_in_scalar(h))
 
     def h3_indexes_are_neighbors(h1, h2):
-        """
-        :returns: boolean
+        """ Returns true if `h1` and `h2` are neighboring cells.
+
+        Returns
+        -------
+        bool
```
isaacbrodsky: note there's an inconsistency between using `bool` and `boolean` as the return type in the doc strings.

src/h3/api/_api_template.py
ajfriend: Yeah, I will before landing. These are still draft notes :)

src/h3/api/_api_template.py
ajfriend: Everything should be `bool` now, with the recent diffs.

src/h3/api/_api_template.py
ajfriend: updating!

src/h3/api/_api_template.py
ajfriend: updating!

src/h3/api/_api_template.py
ajfriend: I've updated the description. Let me know what you think!

src/h3/api/_api_template.py
ajfriend: corrected

src/h3/api/_api_template.py
ajfriend: updated

src/h3/api/_api_template.py
ajfriend: note added

src/h3/api/_api_template.py
ajfriend: agreed!

src/h3/api/_api_template.py
ajfriend: got it, thanks

ajfriend: I'll add a test for this.

src/h3/api/_api_template.py
ajfriend: updating

src/h3/api/_api_template.py
ajfriend: @isaacbrodsky, by "base cell number", do you mean the number of the face of the icosahedron?

ajfriend: The terminology is a little confusing, because here (https://github.com/uber/h3/blob/master/src/h3lib/include/h3api.h.in#L315-L321), "base cell" seems to mean the face integer number.

But here, (https://github.com/uber/h3/blob/master/src/h3lib/include/h3api.h.in#L285-L294), the comment seems to suggest that "base cell" is equivalent to "resolution 0 index".

ajfriend: OK, I read the docs a bit closer. I see the difference now. This returns a base cell number from 0 to 121, and the base cell H3 indices would have that number encoded in the H3Index for a cell with resolution zero.

What would be good terminology to distinguish between the 0--121 "base cell" and the H3Index "base cell" representations?

ajfriend: Updated the description. Let me know what you think.

src/h3/api/_api_template.py
```diff
@@ -19,17 +55,16 @@ def _api_functions(
         _globals,
 ):
     def versions():
-        """ Return version numbers for both the Python library
-        and the underlying H3 C library.
+        """
+        Return version numbers for both the Python anc C libraries.
```
isaacbrodsky: typo: anc -> and

src/h3/api/_api_template.py
```diff
@@ -136,24 +264,91 @@ def h3_distance(h1, h2):
...
+        Return tuple of lat/lng pairs describing the cell boundary.
+
+        Parameters
+        ----------
+        h : H3Cell
+        geo_json : bool, optional
+            If `True`, return output in GeoJson format:
+            lng/lat pairs (opposite order), and
+            have the last pair be the same as the first.
+            If `False`, return lat/lng pairs, with the last
```
isaacbrodsky: "If `False` (default)"

src/h3/api/_api_template.py
```diff
@@ -218,33 +485,123 @@ def polyfill_geojson(geojson, res):
...
+        Returns
+        -------
+        int
         """
         return _cy.get_base_cell(_in_scalar(h))
 
     def h3_indexes_are_neighbors(h1, h2):
         """
-        :returns: boolean
+        Returns true if `h1` and `h2` are neighboring cells.
```
isaacbrodsky: nit: "Returns `True`"

src/h3/api/_api_template.py
```diff
@@ -218,33 +485,123 @@ def polyfill_geojson(geojson, res):
...
     def get_h3_unidirectional_edge(origin, destination):
+        """
+        Create an H3 Index denoting a unidirectional edge.
+
+        The edge is constructed from neighboring cells `origin` and
+        `destination`.
+
+        Parameters
+        ----------
+        h1 : H3Cell
```
isaacbrodsky: nit: rename parameters in the docstring to match the more descriptive names

PR #93, opened 2020-01-20
Changes to build wheel for Windows 10 64bit

nastasi-oq:
To be able to build ``h3`` wheel I installed on a new Windows 10 64 bit machine:
 * git for windows
 * git bash for windows
 * python 3.7 for windows
 * cmake for windows
 * Microsoft Visual Studio Community Edition (configuration export attached)

Made the 2 changes in the PR.

Run the command: ``python setup.py bdist_wheel``

I hope it will help you.

[dot_vsconfig.zip](https://github.com/uber/h3-py/files/4085229/dot_vsconfig.zip)

### DIFF
--- a/.install.sh
+++ b/.install.sh
@@ -37,7 +37,9 @@ if [ "Windows_NT" != "$OS" ]; then
 else
   # Assumed to be Windows, default to x86
   if [[ "True" == "$IS_64BITS" ]]; then
-    cmake . -DENABLE_FORMAT=OFF -DBUILD_SHARED_LIBS=ON -G "Visual Studio 16 2019"
+    # to build on win 10 with VS 16 use this command:
+    # cmake . -DENABLE_FORMAT=OFF -DBUILD_SHARED_LIBS=ON -G "Visual Studio 16 2019"
+    cmake . -DENABLE_FORMAT=OFF -DBUILD_SHARED_LIBS=ON -G "Visual Studio 14 Win64"
   else
     cmake . -DENABLE_FORMAT=OFF -DBUILD_SHARED_LIBS=ON
   fi
--- a/.install.sh
+++ b/.install.sh
@@ -38,8 +38,7 @@ else
   # Assumed to be Windows, default to x86
   if [[ "True" == "$IS_64BITS" ]]; then
     # to build on win 10 with VS 16 use this command:
-    # cmake . -DENABLE_FORMAT=OFF -DBUILD_SHARED_LIBS=ON -G "Visual Studio 16 2019"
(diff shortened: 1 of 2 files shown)


.install.sh
```diff
@@ -37,6 +37,8 @@ if [ "Windows_NT" != "$OS" ]; then
 else
   # Assumed to be Windows, default to x86
   if [[ "True" == "$IS_64BITS" ]]; then
+    # to build on win 10 with VS 16 use this command:
+    # cmake . -DENABLE_FORMAT=OFF -DBUILD_SHARED_LIBS=ON -G "Visual Studio 16 2019"
     cmake . -DENABLE_FORMAT=OFF -DBUILD_SHARED_LIBS=ON -G "Visual Studio 14 Win64"
```
isaacbrodsky: I think we can use `-A x64` to avoid needing to set a specific version of Visual Studio. (There's an example of this in https://github.com/uber/h3-java/blob/master/src/main/c/h3-java/build-h3-windows.ps1#L69)

nastasi-oq: @isaacbrodsky on my windows setup seems to work properly, I'm adding the change to the PR

.install.sh
```diff
@@ -37,7 +37,8 @@ if [ "Windows_NT" != "$OS" ]; then
 else
   # Assumed to be Windows, default to x86
   if [[ "True" == "$IS_64BITS" ]]; then
-    cmake . -DENABLE_FORMAT=OFF -DBUILD_SHARED_LIBS=ON -G "Visual Studio 14 Win64"
+    # to build on win 10 with VS 16 use this command:
```
isaacbrodsky: Not sure this comment is still needed since this isn't just for Windows 10, but also other versions.

.install.sh
isaacbrodsky: This change looks good to me, thanks!

.install.sh
dfellis: While that's true, Win 7 just entered EOL so technically the only version of Windows we need to target is 10.

isaacbrodsky: There's still Windows 8.1 ;)

dfellis: Oh. Right. Does anybody actually use that, though? :)

nastasi-oq: @isaacbrodsky, @dfellis : no strong opinion about it, just tell me what you (and the rest of the team) decide.

dfellis: I am not going to block the PR on that. If someone on the team feels strongly about it, they can change the comment when they do the patch bump to release it, I think.

PR #95, opened 2020-02-09
Add `get_pentagons` function

ajfriend:


src/h3/_cy/cells.pyx
```diff
@@ -267,3 +267,15 @@ cpdef H3int[:] line(H3int start, H3int end):
 
 cpdef bool is_res_class_iii(H3int h):
     return h3lib.h3IsResClassIII(h) == 1
+
+
+cpdef H3int[:] get_pentagons(int res):
```
nrabinowitz: Should this diverge from the core H3 library name (which I assume would be `get_pentagon_indexes`)? Obviously we're planning to rename these functions anyway, but it seems like this is jumping the gun on that - might be more consistent to keep the v3 names for now.

ajfriend: fair enough!

PR #99, opened 2020-04-25
The big Cython move

ajfriend:


CHANGELOG.md
```diff
@@ -6,14 +6,12 @@ numbers of H3-Py are the major and minor version of the bound core library,
...
+We use [this changelog structure](http://keepachangelog.com/).
+
 Because H3-Py is versioned in lockstep with the H3 core library, please
 avoid adding features or APIs which do not map onto the
 [H3 core API](https://uber.github.io/h3/#/documentation/api-reference/).
 
-# Change Log
-
-All notable changes to this project will be documented in this file.
-This project adheres to [Semantic Versioning](http://semver.org/) and uses [this changelog structure](http://keepachangelog.com/).
```
isaacbrodsky, line 15: Add a changelog entry in this PR? Or when publishing after this PR?

.travis.yml
```diff
@@ -3,10 +3,14 @@ cache: pip
 os:
 - linux
 python:
-  - "2.7"
-  - "3.6"
+  - 2.7
```
isaacbrodsky, line 6: We're supporting Python 2.7?

appveyor.yml
```diff
@@ -21,10 +19,17 @@ platform:
...
-  - fab --version
-  - fab test
+  - pytest tests/test_basic_int.py
+  - pytest tests/test_basic_str.py
+  - pytest tests/test_cells_and_edges.py
+  - pytest tests/test_h3.py
+  - pytest tests/test_memview_int.py
+  - pytest tests/test_numpy_int.py
+  - pytest tests/test_polyfill.py
+  - pytest tests/test_to_multipoly.py
```
isaacbrodsky, line 35: Can this be `pytest tests/*.py`?

setup.py
```diff
@@ -1,65 +1,26 @@
...
-        ['out/*.dylib' if platform.system() == 'Darwin' else (
-            'out/*.dll' if platform.system() == 'Windows' else
-            'out/*.so.*')]
-    },
-    license='Apache License 2.0',
-    distclass=BinaryDistribution)
+    name = 'h3',
+    version = about['__version__'],
+    description = about['__description__'],
+    long_description = 'long_description',
```
isaacbrodsky, line 15: `'long_description'` -> `about['__description__']`? Or add a long description text?

src/h3/_cy/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+
+macro(add_cython_file filename)
+    add_cython_target(${filename} C PY3)
+    add_library(${filename} MODULE ${filename} ${LIB_SOURCE_FILES} ${CONFIGURED_API_HEADER})
+    python_extension_module(${filename})
+    set_property(TARGET ${filename} PROPERTY C_STANDARD 99)
+    target_compile_definitions(${filename} INTERFACE $<TARGET_PROPERTY:h3,COMPILE_DEFINITIONS>)
+    target_include_directories(${filename} INTERFACE $<TARGET_PROPERTY:h3,INCLUDE_DIRECTORIES>)
```
isaacbrodsky, line 12: I would assume these should be automatic and not needed if you change the target_link_libraries to `target_link_libraries(${filename} PUBLIC h3)`

src/h3/_cy/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+    target_link_libraries(${filename} h3)
+    install(TARGETS ${filename} LIBRARY DESTINATION src/h3/_cy)
+endmacro()
+
+# GLOB pattern is reccomended against
```
isaacbrodsky, line 17: nit typo reccommended => recommended

src/h3/_cy/cells.pyx
```diff
@@ -0,0 +1,281 @@
...
+
+    return h3lib.h3GetResolution(h)
+
+
+cpdef H3int parent(H3int h, res=None) except 1:
+    check_cell(h)
+
+    if res is None:
+        res = resolution(h) - 1
+    # todo: actually, do we want to raise an error if there are no children, or just return an empty set?
```
isaacbrodsky, line 51: todo sounds like it's applicable to the `children` operation - maybe delete?

src/h3/_cy/cells.pyx
```diff
@@ -0,0 +1,281 @@
...
+
+    return mv
+
+
+cpdef H3int[:] _ring_fallback(H3int h, int k):
+    """
+    `ring` tries to call `h3lib.hexRing` first; if that fails, we call
+    this function, which relies on `h3lib.kRingDistances`.
+
+    Failures for `h3lib.hexRing` happen when that alg runs into a pentagon.
```
isaacbrodsky, line 88: maybe expand `alg` to `algo`/`algorithm`/`func`/`function` - I don't feel like I usually see algorithm abbreviated that way

src/h3/_cy/cells.pyx
```diff
@@ -0,0 +1,281 @@
...
+
+    return mv
+
+cpdef H3int[:] ring(H3int h, int k):
+    """ Return cells at grid distance `== k` from `h`.
+    Collection is "hollow" for k >= 1.
+    """
+    check_cell(h)
+
+    n = 6*k if k > 0 else 1
```
isaacbrodsky, line 120: do we need a check for k < 0?

src/h3/_cy/cells.pyx
```diff
@@ -0,0 +1,281 @@
...
+    mv = create_mv(ptr, n)
+
+    return mv
+
+
+
+cpdef H3int[:] compact(const H3int[:] hu):
+    # todo: the Clib can handle 0-len arrays because it **avoids**
+    # dereferencing the pointer, but Cython's syntax of
+    # `&hu[0]` **requires** a dereference. For Cython, checking for array
```
isaacbrodsky, line 159: I'm still surprised this requires a dereference, in C it would be written the same but I believe not actually cause a dereference.

src/h3/_cy/geo.pyx
```diff
@@ -0,0 +1,270 @@
...
+    lnglat_order : bool
+        If True, assume coordinate pairs like (lng, lat)
+        If False, assume coordinate pairs like (lat, lng)
+    """
+    cdef:
+        h3lib.Geofence gf
+
+    gf.numVerts = len(geos)
+
+    gf.verts = <h3lib.GeoCoord*> stdlib.calloc(gf.numVerts, sizeof(h3lib.GeoCoord))
```
isaacbrodsky, line 71: May want to note in this function's documentation the need for freeing.

src/h3/_cy/geo.pyx
```diff
@@ -0,0 +1,270 @@
...
+    # todo: move this verts transform into the GeoBoundary object
+    verts = tuple(
+        coord2deg(gb.verts[i])
+        for i in range(gb.num_verts)
+    )
+
+    if geo_json:
+        #lat/lng -> lng/lat and last point same as first
+        verts = tuple(tuple(reversed(v)) for v in verts)
+        verts += (verts[0],)
```
isaacbrodsky, line 268: If there's not a performance difference, could we use `tuple([verts[0]])`? This use of the comma operator seems somewhat tricky if you're not familiar with what it's doing.

src/h3/_cy/util.pyx
```diff
@@ -0,0 +1,162 @@
...
+class H3ResolutionError(H3ValueError):
+    pass
+
+cdef check_cell(H3int h):
+    """ Check if valid H3 "cell" (hexagon or pentagon).
+
+    Does not check if a valid H3 edge, for example.
+    """
+    if h3IsValid(h) == 0:
+        raise H3CellError('Integer is not a valid H3 cell: {}'.format(hex(h)))
```
isaacbrodsky, line 52: Want to use `int2hex` here? Probably doesn't matter since this is an error message.

tests/test_basic_int.py
```diff
@@ -0,0 +1,27 @@
+import h3.api.basic_int as h3
+
+
+def test1():
+    assert h3.geo_to_h3(37.7752702151959, -122.418307270836, 9) == 617700169958293503
```
isaacbrodsky, line 5: In the future could we write test indexes in hexadecimal? I think we're all used to reading the indexes in that format.

.travis.yml
ajfriend: Yes. But basically at zero cost to us; Cython handles it all automatically. We can drop it in a future release, or say it is "not officially supported".

But also, I thought we decided that we wanted Python 2 compatibility (well, maybe its been about a year ago now...).

appveyor.yml
ajduberstein: Ideally there should be a pytest.ini file and this is just a single `pytest` command. Example from pydeck [here](https://github.com/visgl/deck.gl/blob/master/bindings/pydeck/pytest.ini).

setup.py
ajduberstein: Ideally this is your README, since it's what appears on the pypi page for the package. Pydeck as an [example](https://github.com/visgl/deck.gl/blob/master/bindings/pydeck/setup.py#L182).

.flake8
```diff
@@ -0,0 +1,14 @@
+# https://lintlyci.github.io/Flake8Rules/
+[flake8]
+max-line-length = 88
+include =
+    #W292 No newline at end of file
+    W292,
+ignore =
+    # E251 unexpected spaces around keyword / parameter equals
+    E251,
+    # sometimes I just want to line up decimal points...
```
ajduberstein, line 10: We agree :)

tests/test_basic_int.py
```diff
@@ -0,0 +1,27 @@
+import h3.api.basic_int as h3
+
+
+def test1():
```
ajduberstein, line 4: I'd bias toward a more descriptive test name, like `test_geo_to_h3`

I'd also group your tests by the function being tested rather than input type being tested. Might be nice to use a [test table](https://github.com/golang/go/wiki/TableDrivenTests) for that

tests/test_basic_int.py
ajfriend: In general, that's what I'm doing in the test suite. This test is here to specifically test the `int64` interface, so I think it's reasonable to have it in decimal.

ajfriend: Also, just updated the test to use *both*  `617700169958293503` and `0x8928308280fffff` to try and make the equivalence apparent.

tests/test_basic_int.py
ajfriend: Added some better names. :)

Most of the testing is through the `str` interface. Since the underlying code paths are the same, I just have this test here to check that the basic conversion functionality works, which should imply that all the other tests will work.

It's true that table testing would be more robust. I'll add a task for us to take a look at this in the future.

ajfriend: issue: https://github.com/uber/h3-py/issues/102

setup.py
ajfriend: updated!

src/h3/_cy/geo.pyx
ajfriend: I'd actually say that the `(verts[0],)` is [pretty idiomatic python](https://stackoverflow.com/questions/1380860/add-variables-to-tuple).

That said, I did refactor a bit. Let me know what you think.

src/h3/_cy/util.pyx
ajfriend: Well, I wanted to use the same `check_cell()` for all interfaces (and not have to repeat code). To do that, this function happens at the C/Cython level, and doesn't know if it is being called by an interface that represents `H3Index`s as `str`s or `int`s.

I figured I basically had these three options:

Screen Shot 2020-05-09 at 2 59 55 PM

Of those, the `hex()` option seemed the most informative: inform users that this **integer representation** of their input is incorrect, but with the hex representation being more informative to them if they are using a `str` interface.

Does that seem reasonable?

Alternatively, I guess we could use `int2hex()` and have the output string prepend `0x`. But not sure if there's a benefit there.

dfellis: The string representation of the H3 indexes is hexadecimal without a prefixed `0x` though.

ajfriend: Also added a note in the function docstring.

ajfriend: > The string representation of the H3 indexes is hexadecimal without a prefixed `0x` though.

Yes, my thinking is that `0x...` is more "correct" in terms of the language and integer representation.

dfellis: Why would you deviate from what every other binding and the C code itself is doing?

src/h3/_cy/geo.pyx
ajfriend: Good point. Added.

CHANGELOG.md
ajfriend: updated!

src/h3/_cy/CMakeLists.txt
ajfriend: that's not a nit; that's just plain wrong!

fixed!

src/h3/_cy/util.pyx
ajfriend: I think one of the points of this re-write was to deviate from the other bindings. ;)
At least, in the fact that we're trying to offer `uint64` interfaces as well as the otherwise standard string interface.

In that context, my thinking is that the fundamentally "correct" and unambiguous representation of an `H3Index` is its `uint64` representation. I realize that the `str` representation is the one people are used to, which is why I was trying to strike a balance in this error message. 

It's **correct** in the sense that the error is saying "The integer `0x...` that i received is not a valid H3Index", but it is **familiar** in the sense that a user sees the hex representation.

Suppose, for example, a user comes to this library at some point in the future and only uses the `numpy_int` interface. When they see an error that hex `'f'` is invalid, they might think "What!? I never put in 'f'!", but if they see that hex `0xf` is invalid, they have a better idea that it is talking about an int.

ajfriend: I guess an alternative approach would be to have custom error message formatting for each interface. Maybe that's a good future issue to tackle?

appveyor.yml
ajfriend: IIRC, I was hoping a single `pytest tests/*` would work, but the test discovery seemed to mess up on appveryor for some reason, which is what led me to listing out each file separately.

I'll give it a shot again, otherwise I'll turn it into an issue to fix in the future.

src/h3/_cy/cells.pyx
ajfriend: replaced with "algorithm".

src/h3/_cy/cells.pyx
ajfriend: good catch! fixed.

src/h3/_cy/cells.pyx
ajfriend: nice catch! I added 2 tests, a new error type, and some checks

src/h3/_cy/CMakeLists.txt
ajfriend: I have no idea how CMake works, so I just tried your suggestion. I got the following error:

Screen Shot 2020-05-09 at 5 03 47 PM

I'll turn this into an issue, and we can revisit after landing this to master?

ajfriend: https://github.com/uber/h3-py/issues/103

appveyor.yml
ajfriend: Yeah, with `pytest tests/*.py` I get the following error:

Screen Shot 2020-05-09 at 6 35 02 PM

ajfriend: I'm not sure what appveyor wants, so it gets an issue: https://github.com/uber/h3-py/issues/104

src/h3/_cy/cells.pyx
ajfriend: I agree this feels awkward and unnecessary, but I couldn't find a better way around it earlier. Let's kick this to an issue and try to figure it out later?

ajfriend: issue: https://github.com/uber/h3-py/issues/105

Issue #104, opened 2020-05-10
Figure out appveyor for testing


ajfriend:
See: https://github.com/uber/h3-py/pull/99#discussion_r415237807

Using `pytest tests/*.py`  in `appveryor.yml` doesn't seem to work, so we have to list each test file independently, which is bad.

ajfriend:
Clsoed with #111

PR #111, opened 2020-05-16
Use pyproject.toml

ajfriend:
`pyproject.toml` notifies `pip` (or similar tools) what is required to build the project. This removes the need for the user to `pip install scikit-build` ahead of time (and perhaps also `cmake` and `ninja`?).

For reference, see:

- https://snarky.ca/what-the-heck-is-pyproject-toml/
- https://scikit-build.readthedocs.io/en/latest/usage.html#basic-usage


ajfriend:
Fixes https://github.com/uber/h3-py/issues/104

PR #112, opened 2020-05-17
Github actions for linux and macos

ajfriend:
# Changes

- replaces Travis CI with github actions
- adds MacOS tests with github actions
- code coverage now done with github actions

# Reasons for changes

- Github actions works cross windows/linux/macos
    - we didn't previously have macos building/testing, and this makes it fairly easy
- Having all CI on the same system (GH actions) means we only need to work with one file format for CI
    - also, the MacOS and Linux Actions are exactly the same, so we can (in the future) re-use all the code and combine it into a single file
- Github actions run **much** faster, particularly when compared to Appveyor
- Github has a nice [marketplace of version-controlled actions](https://github.com/marketplace?type=actions)
    - for example, the codecov action: https://github.com/codecov/codecov-action
    - this can help simplify CI scripts, and ensure we're following community-recognized best practices
- [nice integration to help with building and publishing Python wheels to PyPI](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) ([more info here](https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions), and [maybe even an action that already does this for us on linux](https://github.com/marketplace/actions/python-wheels-manylinux-build))
- i've seen other projects use this successfully (especially for building Python C extensions)
    - https://github.com/oxfordcontrol/qdldl-python/blob/master/.github/workflows/build.yml
    - https://github.com/apache/arrow
    - https://github.com/pandas-dev
    - https://github.com/dask/dask

# Applicability to other H3 repos

I'm not super familiar with the CI in other repos, but I would guess that they could also benefit from moving to Github Actions, at least through simplifying/deduping code and speeding up CI times.

# Future work

- build/test on windows with github actions
- build wheels on all platforms for distribution


dfellis:
Might want to delete the Travis CI integration in this PR, too?

ajfriend:
> Might want to delete the Travis CI integration in this PR, too?

Yeah, that's my plan. Still trying to make sure everything works (like coverage) before I rip out Travis CI :)

isaacbrodsky:
Might be good to include some comments on why you're switching CI systems, and whether that's applicable to the other H3 repos.

ajfriend:
> Might be good to include some comments on why you're switching CI systems, and whether that's applicable to the other H3 repos.

@isaacbrodsky, updated in the first comment!

dfellis:
> Might be good to include some comments on why you're switching CI systems, and whether that's applicable to the other H3 repos.

I can answer those. :)

Travis CI is essentially in maintenance mode now and has weird bugs that won't be fixed. It is also likely that Github Actions will kill the company, so moving over without interruption is better for us, and also increases the chances that Travis CI the company will survive by reducing the subsidy of OSS builds they have to manage.

Github Actions is also generally faster and better integrated into Github, so the experience is nicer. They also have Linux, Mac, and Windows support so it should be possible to simplify the CI setup, as mentioned above, to a single config file, which will make ongoing maintenance easier.

Finally, I would say that the answer to your second question is yes. None of the above reasoning is specific to Python.

.travis.yml
```diff
@@ -1,21 +0,0 @@
-language: python
-cache: pip
```
isaacbrodsky, line 2: You might be losing the pip cache, but if the actions are running faster now anyways, probably doesn't matter.

.github/workflows/ci-macos.yml
```diff
@@ -0,0 +1,44 @@
...
+    branches: [ master ]
+
+jobs:
+  tests:
+    name: "Python ${{ matrix.python-version }}"
+    runs-on: macos-latest
+
+    strategy:
+      matrix:
+        python-version: [2.7, 3.6, 3.7, 3.8]
```
isaacbrodsky, line 16: I would make Linux/Mac OSX one file and have the `runs-on` be part of the matrix. (Unless that's not supported or something?)

.github/workflows/ci-linux.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Install dev requirements
+        run: pip install -r requirements-dev.txt
+
+      - name: Lint
+        run: flake8 src/h3 setup.py tests
+
+      - name: Tests
+        run: pytest --cov=h3 --full-trace --cov-report=xml
+
+      - name: Upload coverage to Codecov
```
isaacbrodsky, line 46: Looks like this is ungated, so all the Linux runs report to codecov - not sure that's intended rather than using a single run to generate the coverage

.travis.yml
ajfriend: Yeah, the current CI finishes in a minute or two.

.github/workflows/ci-macos.yml
ajfriend: It is supported, and I do plan on doing it in the future. Since I'm not super experienced with Github actions, I'm just trying to keep things simple for now until I get more comfortable.

Also, when we build the wheels, I'm not sure how much code will be similar, so this saves me from having to split them back up in that case.

.github/workflows/ci-linux.yml
ajfriend: I'm not actually sure what happens with the multiple reports to codecov, other than that it seems to work OK. Is there a downside to having multiple reports?

isaacbrodsky: I don't know what codecov does in that case.

ajfriend: I'll leave it in for now, as we were doing the same thing with Travis CI, and since everything seems to work OK.

I'll revisit later, when I try to clean up and better organize the CI scripts.

PR #115, opened 2020-05-21
Wheels for Mac/Linux

ajfriend:
Build Wheels with `cibuildwheel`: https://github.com/joerick/cibuildwheel

### DIFF
--- a/.github/workflows/wheels-linux.yml
+++ b/.github/workflows/wheels-linux.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Build wheels
         env:
-          CIBW_SKIP: pp* cp27*
+          CIBW_SKIP: pp* cp27* *-manylinux_i686
           CIBW_TEST_REQUIRES: pytest numpy
           CIBW_TEST_COMMAND: pytest {project}/tests
         run: |
--- a/.github/workflows/wheels-linux.yml
+++ b/.github/workflows/wheels-linux.yml
@@ -27,7 +27,7 @@ jobs:
           CIBW_TEST_COMMAND: pytest {project}/tests
         run: |
           pip install --upgrade cibuildwheel pip setuptools wheel
-          cibuildwheel --output-dir wheelhouse
+          python -m cibuildwheel --output-dir wheelhouse
 
       - name: Build sdist
         run: |
--- a/.github/workflows/wheels-linux.yml
+++ b/.github/workflows/wheels-linux.yml
@@ -8,8 +8,8 @@ on:
 
 jobs:
   tests:
-    name: Python Wheels Mac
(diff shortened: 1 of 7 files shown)


.github/workflows/wheels-linux.yml
```diff
@@ -0,0 +1,37 @@
...
+          submodules: recursive
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Build wheels
+        env:
+          CIBW_SKIP: pp* cp27* *-manylinux_i686
+          CIBW_ENVIRONMENT: "CFLAGS='-lrt'"
```
ajfriend, line 26: Any concern that I'm **overwriting** `CFLAGS` rather than **appending** here?

ajfriend:
@bnaul

.github/workflows/wheels-linux.yml
```diff
@@ -0,0 +1,37 @@
+name: wheels-linux
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
isaacbrodsky, line 7: do we want to  build wheels on PRs too? Or just on master?

.github/workflows/wheels-linux.yml
ajfriend: Fixed this, and dropped the need for linking against `rt` by adding this to the cmake file:

```cmake
set(BUILD_BENCHMARKS OFF CACHE BOOL "" FORCE)
set(BUILD_FILTERS OFF CACHE BOOL "" FORCE)
set(BUILD_GENERATORS OFF CACHE BOOL "" FORCE)
set(BUILD_TESTING OFF CACHE BOOL "" FORCE)
```

The benchmarking code was the only thing requiring `rt`, and turning off everything else here helps to reduce build times quite a bit.

ajfriend: @bnaul This should reduce your build times if you do have to compile yourself.

.github/workflows/wheels-linux.yml
isaacbrodsky: we can revisit this later, but maybe add a gate for publishing the artifacts (always build the artifacts though)

ajfriend: We'll probably want to build the wheels in regular testing, but we should probably have some logic so that we only store the artifacts and upload to PyPI when we're actually making a release. Not sure the best way to trigger that tho...

Maybe something like in this project: https://github.com/oxfordcontrol/qdldl-python/blob/master/.github/workflows/build.yml

```yaml
- name: Release to pypi
  if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
```

PR #116, opened 2020-05-22
Windows wheels

ajfriend:
Building Windows wheels for CPython `3.5`, `3.6`, `3.7`, `3.8`.

TODO: The windows build requires `CIBW_BEFORE_BUILD: pip install scikit-build cython cmake` to specify the build dependencies. This is unnecessary on Linux/mac, as it gets picked up automatically from the `pyproject.toml` file. Not clear why this doesn't work on Windows. Follow this issue: https://github.com/joerick/cibuildwheel/issues/350

### DIFF
--- a/.github/workflows/wheels-windows.yml
+++ b/.github/workflows/wheels-windows.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Build wheels
         env:
           CIBW_SKIP: pp* cp27*
-          CIBW_BEFORE_BUILD: pip install scikit-build cmake cython ninja wheel
+          CIBW_BEFORE_BUILD: pip install scikit-build cmake cython ninja wheel setuptools
           CIBW_TEST_REQUIRES: pytest numpy
           CIBW_TEST_COMMAND: pytest {project}/tests
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/wheels-windows.yml
```diff
@@ -0,0 +1,43 @@
...
+          CIBW_TEST_COMMAND: pytest {project}/tests
+        run: |
+          pip install --upgrade cibuildwheel
+          python -m cibuildwheel --output-dir wheelhouse
+
+      # - name: Upload artifacts to github
+      #   uses: actions/upload-artifact@v2
+      #   with:
+      #     name: wheels
+      #     path: ./wheelhouse
```
dfellis, line 43: What is the purpose of building these without uploading them somewhere?

isaacbrodsky: I understand the point here is to test everything and then to enable uploading after the other parts are done.

ajfriend: Yeah. It seemed a bit wasteful to have github store the artifacts after every push. In an upcoming step, we'll figure out how we want to gate the uploads and push to PyPI to only happen for releases.

ajfriend:
Made a ticket for the windows `pyproject.toml` issue: https://github.com/joerick/cibuildwheel/issues/350

PR #117, opened 2020-05-24
Use `setup.py` for optional requirements instead of `requirements-dev.txt`

ajfriend:


.github/workflows/ci-linux.yml
```diff
@@ -32,10 +32,7 @@ jobs:
           pip install --upgrade pip setuptools wheel
 
       - name: Install from source
-        run: pip install .
-
-      - name: Install dev requirements
-        run: pip install -r requirements-dev.txt
+        run: pip install .[all]
```
isaacbrodsky, line 35: Just curious, what happens if I run `pip install .` instead?

ajfriend: `pip install .` installs the minimum requirements. In this case, no other libraries are installed as dependencies. The options are given in the `setup.py` file:

```python
extras_require={
    'numpy': ['numpy'],
    'test': ['pytest', 'pytest-cov', 'flake8'],
    'all': ['numpy', 'pytest', 'pytest-cov', 'flake8'],
},
```

PR #118, opened 2020-05-25
skip 3.5 on windows ci build wheel

ajfriend:
Playing around with this issue: https://github.com/joerick/cibuildwheel/issues/350


henryiii:
Next experiment: can you set `CIBW_DEPENDENCY_VERSIONS` to `latest` instead of the default (`pinned`)? Not likely to be the problem, but might be work a test.

ajfriend:
Weirdly, the workflows aren't getting triggered now...

henryiii:
Github had an outage, maybe it was during that time?

henryiii:
You could bump to cibuildwheel 1.4.2, and see if that triggers a rebuild too.

henryiii:
Looks like it is trying to install dependencies:

```
  Installing build dependencies: started
  Installing build dependencies: still running...
  Installing build dependencies: finished with status 'done'
```

But it's still not finding them in the environment. Can you please set `CIBW_BUILD_VERBOSITY` to 1? That should show the search and links found for the PEP 517 setup. It's *really* verbose, and I which it was quieter, but at least we could see what it's finding and installing.

.github/workflows/tests.yml
```diff
@@ -1,48 +1,48 @@
-name: tests
```
qa8302, line 1: -

ajfriend: ?

ajfriend:
Follow this issue: https://github.com/uber/h3-py/issues/138

PR #122, opened 2020-05-27
try some new wheels logic

ajfriend:


ajfriend:
Landing this to test if the "release" action works now

Issue #121, opened 2020-05-26
Figure out how to automate releases


ajfriend:
- `releases` branch that triggers action?
- want releases on both PyPI and GitHub, how to coordinate?
- https://github.com/actions/create-release

ajfriend:
https://help.github.com/en/actions/reference/events-that-trigger-workflows

ajfriend:
https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions

ajfriend:
- https://dan.yeaw.me/posts/github-actions-automate-your-python-development-workflow/
- https://hynek.me/articles/python-github-actions/

ajfriend:
Closed by #124

PR #124, opened 2020-05-27
Clean up workflows on GitHub actions

ajfriend:
Break the workflows across task instead of platform:

- lint and coverage
- tests
- building wheels

I made some changes in `wheels.yml` in previous PRs which I landed unreviewed, since I was trying to test the `release` event. Please review `wheels.yml` as well.

Also, in `wheels.yml`, I set up some logic so that when we create a GitHub **release**, publishing a PyPI **release** is also triggered automatically.


ajfriend:
If this looks good, I'd say it closes out https://github.com/uber/h3-py/issues/121

PR #126, opened 2020-05-29
Final steps for releasing v3.6.1

ajfriend:


readme.md
```diff
@@ -18,52 +18,174 @@ For API reference, please see the
 [H3 Documentation](https://h3geo.org/).
 
 **NOTE: This version (v.3.6.1) corresponds to the new Cython
-bindings, which will replace the old bindings (v3.4.3).
-This version is not yet published to PyPI.**
+bindings, which have replaced the old bindings (v3.4.3).**
```
nrabinowitz, line 21: Nit: The `v.3.6.1` here is going to get out of date pretty quickly. I'd change to something like:
```
Note: As of v3.6.1, the `h3-py` uses Cython as the underlying binding framework, replacing the older bindings in v3.4.x.
```

readme.md
```diff
@@ -18,52 +18,174 @@ For API reference, please see the
...
 
-TODO: instructions for installing from pre-built wheels
+`pip install h3`
 
+**New in v3.6.1**: We upload pre-built
+[Python Wheels to PyPI](https://pypi.org/project/h3) for Linux/Mac/Windows,
+which should avoid many previous installation issues.
 
-## Install from GitHub
+Try the library with:
```
nrabinowitz, line 32: Nit: I'd usually go with `## Usage` here.

readme.md
```diff
@@ -18,52 +18,174 @@ For API reference, please see the
...
+>>> resolution = 0
+>>> h3.geo_to_h3(lat, lng, resolution)
+'8075fffffffffff'
+```
 
-Then run:
+## Example gallery
 
-`pip install git+https://github.com/uber/h3-py.git`
+Browse [a collection of example notebooks](https://github.com/uber/h3-py-notebooks),
```
nrabinowitz, line 44: Are the notebooks here compatible with the latest bindings?

readme.md
ajfriend: Actually, we probably don't even need that bit. I just took it out.

readme.md
ajfriend: How about just `Usage:`. I don't know if it needs a whole section to itself.

readme.md
ajfriend: I haven't tried, but they should be. These bindings are passing the same tests as the old bindings.

ajfriend: I'll give them a whirl.

readme.md
```diff
@@ -17,53 +17,174 @@ Python bindings for the
...
+{577973680303243263,
+ 578044049047420927,
+ 578677367745019903,
+ 578782920861286399,
+ 579169948954263551}
 ```
 
+### `h3.api.numpy_int`
+
+H3 indexes are represented as `unit64`s, using `numpy.ndarray`
```
dfellis, line 96: Nit: should be `uint64`

readme.md
```diff
@@ -17,53 +17,174 @@ Python bindings for the
...
 
-The optional `h3.api.numpy_int` API requires `numpy`.
-To have `numpy` installed (if it isn't already) along
-with `h3`, run:
 
-`pip install git+https://github.com/uber/h3-py.git[numpy]`
+### `h3.api.memview_int`
 
-or
+H3 indexes are represented as `unit64`s, using Python
```
dfellis, line 122: Same nit.

readme.md
ajfriend: good catch. fixed!

readme.md
ajfriend: good catch. fixed!

readme.md
nrabinowitz: Up to you - it's the first thing I look for in most READMEs, so it doesn't hurt to make it big :)

dfellis: I approved because I assumed @nrabinowitz 's suggestion here was going to be taken. Hard agree on `## Usage`. It's the convention for nearly all GitHub projects.

ajfriend: updated!

readme.md
ajfriend: The basic `Usage.ipynb` works, but the others need to be upgraded for Python 3. I'll make an issue about it for that repo.

ajfriend: Starting a PR: https://github.com/uber/h3-py-notebooks/pull/2

PR #127, opened 2020-05-30
Fix error checking on `hex2int`

ajfriend:
Change from

```python
cpdef H3int hex2int(H3str h) except 0
```
to
```python
cpdef H3int hex2int(H3str h) except? 0
```

The previous version considered **any** return of `0` an error, which led to confusing error messages for the user. This function simply converts; check for a valid cell happens after conversion.

Changing `except 0` to `except? 0` makes it so that Python only **checks** if there is an error when `0` is returned. [We need *some* C-compatible sentinel value to flag actual/potential errors for pure C/Cython functions](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#error-return-values).

Old error message:
Screen Shot 2020-05-30 at 11 46 54 AM

New error message:
Screen Shot 2020-05-30 at 11 44 51 AM

This PR also adds a test to verify that we get a helpful error message in this example.

### DIFF
--- a/src/h3/_cy/util.pxd
+++ b/src/h3/_cy/util.pxd
@@ -1,6 +1,6 @@
 from .h3lib cimport H3int, H3str
 
-cpdef H3int hex2int(H3str h) except 0
+cpdef H3int hex2int(H3str h) except? 0
 cpdef H3str int2hex(H3int x)
 
 cdef H3int* create_ptr(size_t n) except? NULL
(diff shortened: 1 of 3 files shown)


ajfriend:
I'll land this for now, but more than happy to revisit if we have ideas around better error reporting.

Pandas might have it a little easier because it does have typed arrays for columns, in the form of `numpy` arrays, but I'm not sure if that's the error message you were thinking of.

PR #134, opened 2020-06-03
Add missing function bindings

ajfriend:


src/h3/api/_api_template.py
```diff
@@ -754,4 +754,20 @@ def get_pentagon_indexes(resolution):
 
         return _out_unordered(mv)
 
+    def get_res0_indexes():
+        """
+        Return all cells at a given resolution.
```
dfellis, line 759: This only returns the base cells, right?

ajfriend: good catch!

PR #146, opened 2020-07-01
first draft of v4 names

ajfriend:
Experimenting with the [planned v4.0.0 names in this H3 RFC](https://github.com/uber/h3/blob/master/dev-docs/RFCs/v4.0.0/names_for_concepts_types_functions.md).

Try it with something like:

```python
import h3.unstable.v4 as h3

h = h3.point_to_cell(0,0,0)
h3.cell_to_children(h)
h3.get_pentagons(0) < h3.get_res0_cells()
h3.grid_ring(h, k=5)
h3.grid_disk(h, k=3)
```

### DIFF
--- a/src/h3/unstable/v4.py
+++ b/src/h3/unstable/v4.py
@@ -7,7 +7,7 @@
 
 # what if we organized by module
 
-from ..api.basic_str import (   # pragma: no cover
+from ..api.basic_str import (
     compact as
         compact,
 
--- a/src/h3/unstable/v4.py
+++ b/src/h3/unstable/v4.py
@@ -1,57 +1,52 @@
 # flake8: noqa
 
-# todo: what are the pythonic names?
-
-# what if we put this in an object? what would it look like?
-# a temporary API wrapper
-
-# what if we organized by module
-
 from ..api.basic_str import (
+    # todo: implement is_valid_index
+
     compact as
-        compact,
+        compact_cells,
 
     edge_length as
-        hex_edge_length_avg,
+        get_hexagon_edge_length_avg,
 
     geo_to_h3 as
         point_to_cell,
 
     get_destination_h3_index_from_unidirectional_edge as
-        edge_destination,
+        get_directed_edge_destination,
 
     get_h3_indexes_from_unidirectional_edge as
-        directed_edge_cells,
+        directed_edge_to_cells,
 
     get_h3_unidirectional_edge as
         cells_to_directed_edge,
 
     get_h3_unidirectional_edge_boundary as
-        directed_edge_boundary, # what's a boundary?
+        directed_edge_to_boundary,
 
     get_h3_unidirectional_edges_from_hexagon as
         origin_to_directed_edges,
 
     get_origin_h3_index_from_unidirectional_edge as
-        edge_origin,
+        get_directed_edge_origin,
 
     get_pentagon_indexes as
-        pentagons,
+        get_pentagons,
 
     get_res0_indexes as
-        res0_cells,
+        get_res0_cells,
 
     h3_distance as
         grid_distance,
 
     h3_get_base_cell as
-        base_cell_number,
+        get_base_cell_number,
 
     h3_get_faces as
-        faces,
+        get_icosahedron_faces,
 
     h3_get_resolution as
-        resolution,
+        get_resolution,
 
     h3_indexes_are_neighbors as
         are_neighbor_cells,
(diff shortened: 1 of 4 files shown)


src/h3/unstable/v4.py
```diff
@@ -0,0 +1,135 @@
...
+        hex_edge_length_avg,
+
+    geo_to_h3 as
+        point_to_cell,
+
+    get_destination_h3_index_from_unidirectional_edge as
+        edge_destination,
+
+    get_h3_indexes_from_unidirectional_edge as
+        directed_edge_cells,
```
isaacbrodsky, line 24: A little inconsistent that this doesn't have "to" in the name but `cells_to_directed_edge` does. I think I understand why you chose this but it does jump out at me when reading the code.

src/h3/unstable/v4.py
```diff
@@ -0,0 +1,135 @@
...
+        res0_cells,
+
+    h3_distance as
+        grid_distance,
+
+    h3_get_base_cell as
+        base_cell_number,
+
+    h3_get_faces as
+        faces,
```
isaacbrodsky, line 51: This one does seem a little obscure as a function name rather than a method on an index object.

src/h3/unstable/v4.py
```diff
@@ -0,0 +1,135 @@
...
+        hex_area_avg,
+
+    # hex_range as _,
+    # hex_range_distances as _, # not sure we want this one; easy for user to implement
+    # hex_ranges as _,
+
+    hex_ring as
+        grid_ring,
+    hex_ring as
+        ring, # used often enough that i think it deserves an alias
```
isaacbrodsky, line 106: in the future I would suggest making these aliases consistent across bindings. I find it very helpful when switching between different parts of a system to be able to reference the same function names.

isaacbrodsky:
Exciting to see these new function names get tested out in the Python bindings :)

src/h3/unstable/v4.py
```diff
@@ -0,0 +1,135 @@
...
+    polyfill as
+        polygon_to_cells,
+
+    polyfill_geojson as
+        geojson_to_cells, # still need to figure out the final polyfill interface
+
+    #polyfill_polygon as _,
+
+    string_to_h3 as
+        str_to_int,
```
nrabinowitz, line 128: :/ I mean, yes, that's what it does, but if I saw this in isolation I'm not sure I'd expect it to mean "parse a hexadecimal string to a 64-bit int", and when you put it like that it seems odd that the library includes it at all...

ajfriend: But these names are all under the `h3` namespace. It would be used like `h3.str_to_int`, which, I think, makes it clear that this is an H3-specific function.

I don't think `string_to_h3` makes sense either. Would there be a better name?

src/h3/unstable/v4.py
```diff
@@ -0,0 +1,135 @@
...
+        grid_path_cells,
+
+    h3_set_to_multi_polygon as
+        cells_to_multipolygon,
+
+    h3_to_center_child as
+        center_child,
+
+    h3_to_children as
+        children,
```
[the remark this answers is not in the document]
ajfriend: This shouldn't have name collisions, as `children` is behind the `h3` namespace.

You could still do `children = h3.children(index)`.

Does that seem OK?

src/h3/unstable/v4.py
```diff
@@ -0,0 +1,135 @@
...
+        directed_edge_boundary, # what's a boundary?
+
+    get_h3_unidirectional_edges_from_hexagon as
+        origin_to_directed_edges,
+
+    get_origin_h3_index_from_unidirectional_edge as
+        edge_origin,
+
+    get_pentagon_indexes as
+        pentagons,
```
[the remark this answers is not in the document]
ajfriend: Yeah, I think I agree with you on `pengagons` and `faces`. Even `get_faces(cell)` seems a little mysterious. I wonder if something like `get_icosahedon_faces(cell)` would be better (here and in the naming RFC).

src/h3/unstable/v4.py
ajfriend: Agreed. Going with a more verbose name for faces.

ajfriend:
Experimental naming ideas I'll sideline for now: https://gist.github.com/ajfriend/5b1dd78554f606bb307d244c3dacae54

Some other ideas: https://github.com/uber/h3-py/issues/149

ajfriend:
Also added some missing names to the naming RFC: https://github.com/uber/h3/pull/368

PR #147, opened 2020-07-05
Vectorized functions

ajfriend:


ajfriend:
Remove the demo notebook and move it to the demo notebooks repo.

src/h3/_cy/unstable_vect.pyx
```diff
@@ -0,0 +1,59 @@
+cimport h3lib
+from h3lib cimport H3int
+from .util cimport deg2coord
+
+from cython cimport boundscheck, wraparound
+from libc.math cimport sqrt, sin, cos, asin
+
+cdef double haversineDistance(double th1, double ph1, double th2, double ph2) nogil:
```
dfellis, line 8: Does this belong in this binding?

I don't think we provide haversine distance support in any of the other bindings?

ajfriend: I'm considering anything under `h3.unstable` as "not in the binding". There was a request for an optimized `cell_haversine` function, so I'm putting it in to make it easy for them to play with, and evaluate the speedups.

In the short term, this is hard to do anywhere except within the Cython bindings. In the future however, we could consider making it easy for folks to do this *outside* the Cython bindings, by letting them write their own Cython code
which just does `cimport h3` to get access to the C and Cython code in these bindings. Here's an issue around it: https://github.com/uber/h3-py/issues/152

I'd say that's probably the way to go in the future.

src/h3/_cy/h3lib.pxd
```diff
@@ -48,9 +48,9 @@ cdef extern from "h3api.h":
         LinkedGeoLoop *_data_last "last"  # not needed in Cython bindings
         LinkedGeoPolygon *next
 
-    H3Index geoToH3(const GeoCoord *g, int res)
+    H3Index geoToH3(const GeoCoord *g, int res) nogil
```
isaacbrodsky, line 51: How do you determine which functions get `nogil`? Is it just the ones you want to optimize right now?

ajfriend: I just added it to the functions I needed it for in this PR. We could probably add `nogil` to each of the functions in the API, as the `nogil` tag just promises that the function won't manipulate any Python objects, which is true for everything in the `h3-c` API.

I'll probably do a more thorough job of this when I tackle https://github.com/uber/h3-py/issues/140

src/h3/_cy/unstable_vect.pyx
kylebarron: It's [pretty easy](https://stackoverflow.com/a/29546836) to make a vectorized `haversine` in numpy only. So for the future if/when vectorized functions make it out of `unstable`, it might be ideal to only make `h3_to_geo` vectorized, then the end user can have a vectorized pipeline without touching cython themselves:

```py
# a, b arrays of hex ids
a_geo = h3_to_geo(a)
b_geo = h3_to_geo(b)
dist = haversine(a, b)
```

ajfriend: We will expose a haversine function when https://github.com/uber/h3/pull/377 lands:

https://github.com/uber/h3/blob/930cb1be6efe88ef397fb443027a2a86c0616c88/src/h3lib/include/h3api.h.in#L265-L269

It would be easy to add that to the `h3-py` "vectorized" API, whenever we figure out the best way to do that. I'd be interesting to compare it to the "numpy-vectorized" version you have above!

PR #155, opened 2020-07-18
local ij conversion functions

ajfriend:


tests/test_cells_and_edges.py
```diff
@@ -568,3 +568,49 @@ def test_get_faces():
...
+
+
+def test_to_local_ij_self():
+    h = h3.geo_to_h3(0,0,9)
+    out = h3.experimental_h3_to_local_ij(h, h)
+
+    assert out == (-858, -2766)
+
+    # weird, shouldn't it be:
+    # assert out == (0,0)
```
ajfriend, line 616: I'm a little confused by this behavior. Why wouldn't the local (i,j) from a cell to itself be `(0,0)`?

Maybe I'm misunderstanding the functions. @isaacbrodsky, can you shed some light here?

tests/test_cells_and_edges.py
```diff
@@ -568,3 +568,49 @@ def test_get_faces():
...
+
+
+def test_to_local_ij_self():
+    h = h3.geo_to_h3(0, 0, 9)
+    out = h3.experimental_h3_to_local_ij(h, h)
+
+    assert out == (-858, -2766)
+
+    # weird, shouldn't it be:
+    # assert out == (0,0)
```
isaacbrodsky, line 616: This is expected behavior. The `origin` cell does not define (0, 0) for the IJ coordinate space. (0, 0) refers to the center of the base cell containing `origin` at the resolution of `origin`. (This is an implementation detail.) Subtracting the IJ coordinates of `origin` from every cell should get you the property of (0, 0) being the origin.

ajfriend: Got it. Thanks!

PR #156, opened 2020-07-20
Fix walkthrough notebook broken link

alejandrohall:


ajfriend:
Thanks for this! 

Those notebooks *should* work, but haven't been seriously updated in a while, so they could probably use some love. If you see any issues with them, please do let us know! Or, even better, submit a PR ;)

PR #161, opened 2020-08-01
Area functions bug

ajfriend:


ajfriend:
Close because it is so old.

PR #166, opened 2020-09-03
Add vectorized h3_to_parent and h3_get_resolution

kylebarron:
This PR adds a simple vectorized `h3_to_parent`, along the lines of #147 . I compared this to a simple loop that repeatedly calls `h3.h3_to_parent` and found a 160x speedup over pure Python and 100x faster than `np.vectorize`.

I'd love to add a vectorized complement to every h3 API function, but that would create a lot of duplication, as presumably every function would need a for loop in Cython, so you couldn't use the `_api_template` setup to add vectorized functions.

```py
import h3.api.numpy_int as h3
import h3.unstable.vect as vect
import numpy as np

N = 100_000
K = 10

lats, longs = np.random.uniform(0, 90, N), np.random.uniform(0, 90, N)
h3_index = vect.geo_to_h3(lats, longs, 9)

%%timeit
out_loop = []
for i in range(len(h3_index)):
    out_loop.append(h3.h3_to_parent(h3_index[i], 8))
# 67.6 ms ± 2.06 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

vec_fn = np.vectorize(h3.h3_to_parent)
%timeit vec_fn(h3_index, 8)
# 42.6 ms ± 1.23 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit out_vect = vect.h3_to_parent(h3_index, 8)
# 423 µs ± 15.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

np.array_equal(out_loop, out_vect)
# True
```

cc @isaacbrodsky

### DIFF
--- a/src/h3/_cy/unstable_vect.pyx
+++ b/src/h3/_cy/unstable_vect.pyx
@@ -64,7 +64,7 @@ cpdef void geo_to_h3_vect(
 @wraparound(False)
 cpdef void h3_to_parent_vect(
     const H3int[:] h,
-    unsigned char[:] res,
+    int[:] res,
     H3int[:] out
 ) nogil:
 
@@ -79,7 +79,7 @@ cpdef void h3_to_parent_vect(
 @wraparound(False)
 cpdef void h3_get_resolution_vect(
     const H3int[:] h,
-    unsigned char[:] out,
+    int[:] out,
 ) nogil:
 
     cdef Py_ssize_t i
--- a/src/h3/_cy/unstable_vect.pyx
+++ b/src/h3/_cy/unstable_vect.pyx
@@ -1,5 +1,5 @@
 cimport h3lib
-from h3lib cimport H3int, H3Index
+from h3lib cimport H3int
 from .util cimport deg2coord
 
 from cython cimport boundscheck, wraparound
@@ -42,6 +42,7 @@ cpdef void haversine_vect(
                 p2.lat, p2.lng
             )
 
+
 @boundscheck(False)
 @wraparound(False)
 cpdef void geo_to_h3_vect(
@@ -58,14 +59,31 @@ cpdef void geo_to_h3_vect(
             c = deg2coord(lat[i], lng[i])
             out[i] = h3lib.geoToH3(&c, res)
 
+
 @boundscheck(False)
 @wraparound(False)
 cpdef void h3_to_parent_vect(
-    const H3Index[:] h,
-    int res,
-    H3Index[:] out
+    const H3int[:] h,
+    unsigned char[:] res,
+    H3int[:] out
 ) nogil:
+
+    cdef Py_ssize_t i
 
     with nogil:
         for i in range(len(h)):
-            out[i] = h3lib.h3ToParent(h[i], res)
+            out[i] = h3lib.h3ToParent(h[i], res[i])
+
+
+@boundscheck(False)
+@wraparound(False)
+cpdef void h3_get_resolution_vect(
+    const H3int[:] h,
+    unsigned char[:] out,
+) nogil:
+
+    cdef Py_ssize_t i
+
+    with nogil:
+        for i in range(len(h)):
+            out[i] = h3lib.h3GetResolution(h[i])
--- a/src/h3/_cy/unstable_vect.pyx
+++ b/src/h3/_cy/unstable_vect.pyx
@@ -1,5 +1,5 @@
 cimport h3lib
-from h3lib cimport H3int
+from h3lib cimport H3int, H3Index
 from .util cimport deg2coord
 
 from cython cimport boundscheck, wraparound
@@ -57,3 +57,17 @@ cpdef void geo_to_h3_vect(
         for i in range(len(lat)):
             c = deg2coord(lat[i], lng[i])
             out[i] = h3lib.geoToH3(&c, res)
+
+@boundscheck(False)
+@wraparound(False)
+cpdef void h3_to_parent_vect(
+    const H3Index[:] h,
+    int res,
+    H3Index[:] out
+):
+
+    # cdef H3Index c
+
+    for i in range(len(h)):
+        # c = h[i]
+        out[i] = h3lib.h3ToParent(h[i], res)
--- a/src/h3/_cy/unstable_vect.pyx
+++ b/src/h3/_cy/unstable_vect.pyx
@@ -64,7 +64,8 @@ cpdef void h3_to_parent_vect(
     const H3Index[:] h,
     int res,
     H3Index[:] out
-):
+) nogil:
 
-    for i in range(len(h)):
-        out[i] = h3lib.h3ToParent(h[i], res)
+    with nogil:
+        for i in range(len(h)):
+            out[i] = h3lib.h3ToParent(h[i], res)
--- a/src/h3/_cy/unstable_vect.pyx
+++ b/src/h3/_cy/unstable_vect.pyx
@@ -66,8 +66,5 @@ cpdef void h3_to_parent_vect(
     H3Index[:] out
 ):
 
-    # cdef H3Index c
-
     for i in range(len(h)):
-        # c = h[i]
         out[i] = h3lib.h3ToParent(h[i], res)
--- a/src/h3/unstable/vect.py
+++ b/src/h3/unstable/vect.py
@@ -18,6 +18,7 @@ def h3_to_parent(h, res=None):
     -------
     array of H3Cells
     """
+    h = np.array(h, dtype=np.uint64)
     out = np.zeros(len(h), dtype=np.uint64)
 
     if res is None:
@@ -42,6 +43,7 @@ def h3_get_resolution(h):
     -------
     array of uint8
     """
+    h = np.array(h, dtype=np.uint64)
     out = np.zeros(len(h), dtype=np.uint8)
 
     _vect.h3_get_resolution_vect(h, out)
@@ -64,6 +66,9 @@ def geo_to_h3(lats, lngs, res):
     -------
     array of H3Cells
     """
+    lats = np.array(lats, dtype=np.float64)
+    lngs = np.array(lngs, dtype=np.float64)
+
     assert len(lats) == len(lngs)
 
     out = np.zeros(len(lats), dtype='uint64')
@@ -88,6 +93,8 @@ def cell_haversine(a, b):
     float
         Haversine distance in kilometers
     """
+    a = np.array(a, dtype=np.uint64)
+    b = np.array(b, dtype=np.uint64)
 
     assert len(a) == len(b)
 
--- a/src/h3/unstable/vect.py
+++ b/src/h3/unstable/vect.py
@@ -41,7 +41,7 @@ def h3_get_resolution(h):
 
     Returns
     -------
-    array of uint8
+    array of int
     """
     h = np.array(h, dtype=np.uint64)
     out = np.zeros(len(h), dtype=np.intc)
--- a/src/h3/unstable/vect.py
+++ b/src/h3/unstable/vect.py
@@ -24,7 +24,7 @@ def h3_to_parent(h, res=None):
     if res is None:
         res = h3_get_resolution(h) - 1
     else:
-        res = np.full(h.shape, res, dtype=np.int16)
+        res = np.full(h.shape, res, dtype=np.intc)
 
     _vect.h3_to_parent_vect(h, res, out)
 
@@ -44,7 +44,7 @@ def h3_get_resolution(h):
     array of uint8
     """
     h = np.array(h, dtype=np.uint64)
-    out = np.zeros(len(h), dtype=np.int16)
+    out = np.zeros(len(h), dtype=np.intc)
 
     _vect.h3_get_resolution_vect(h, out)
 
--- a/src/h3/unstable/vect.py
+++ b/src/h3/unstable/vect.py
@@ -2,6 +2,7 @@ from .._cy import unstable_vect as _vect
 
 import numpy as np
 
+
 def h3_to_parent(h, res):
     """
     Get parent of arrays of cells
@@ -22,6 +23,7 @@ def h3_to_parent(h, res):
     _vect.h3_to_parent_vect(h, res, out)
 
     return out
+
 
 def geo_to_h3(lats, lngs, res):
     """
--- a/src/h3/unstable/vect.py
+++ b/src/h3/unstable/vect.py
@@ -3,24 +3,48 @@ from .._cy import unstable_vect as _vect
 import numpy as np
 
 
-def h3_to_parent(h, res):
+def h3_to_parent(h, res=None):
     """
-    Get parent of arrays of cells
+    Get parent of arrays of cells.
 
     Parameters
     ----------
     h : array of H3Cells
-
-    res: int
-        Resolution for output cells.
+    res: int or None, optional
+        The resolution for the parent
+        If `None`, then `res = resolution(h) - 1`
 
     Returns
     -------
     array of H3Cells
     """
-    out = np.zeros(len(h), dtype='uint64')
+    out = np.zeros(len(h), dtype=np.uint64)
+
+    if res is None:
+        res = h3_get_resolution(h) - 1
+    else:
+        res = np.full(h.shape, res, dtype=np.uint8)
 
     _vect.h3_to_parent_vect(h, res, out)
+
+    return out
+
+
+def h3_get_resolution(h):
+    """
+    Return the resolution of an array of H3 cells.
+
+    Parameters
+    ----------
+    h : H3Cell
+
+    Returns
+    -------
+    array of uint8
+    """
+    out = np.zeros(len(h), dtype=np.uint8)
+
+    _vect.h3_get_resolution_vect(h, out)
 
     return out
 
--- a/src/h3/unstable/vect.py
+++ b/src/h3/unstable/vect.py
@@ -24,7 +24,7 @@ def h3_to_parent(h, res=None):
     if res is None:
         res = h3_get_resolution(h) - 1
     else:
-        res = np.full(h.shape, res, dtype=np.uint8)
+        res = np.full(h.shape, res, dtype=np.int16)
 
     _vect.h3_to_parent_vect(h, res, out)
 
@@ -44,7 +44,7 @@ def h3_get_resolution(h):
     array of uint8
     """
     h = np.array(h, dtype=np.uint64)
-    out = np.zeros(len(h), dtype=np.uint8)
+    out = np.zeros(len(h), dtype=np.int16)
 
     _vect.h3_get_resolution_vect(h, out)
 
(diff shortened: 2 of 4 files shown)


src/h3/_cy/unstable_vect.pyx
```diff
@@ -57,3 +57,17 @@ cpdef void geo_to_h3_vect(
...
             c = deg2coord(lat[i], lng[i])
             out[i] = h3lib.geoToH3(&c, res)
+
+@boundscheck(False)
+@wraparound(False)
+cpdef void h3_to_parent_vect(
+    const H3Index[:] h,
+    int res,
+    H3Index[:] out
+):
```
kylebarron, line 67: I'm not a pro at Cython, nor do I understand the H3 C API too well. When I try setting `nogil` like so:
```py
@boundscheck(False)
@wraparound(False)
cpdef void h3_to_parent_vect(
    const H3Index[:] h,
    int res,
    H3Index[:] out
) nogil:

    with nogil:
        for i in range(len(h)):
            out[i] = h3lib.h3ToParent(h[i], res)
```

I get a compilation error. Maybe I'm doing something wrong, but regardless vectorized with GIL is still a huge improvement over non-vectorized.
```
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
        H3Index[:] out
    ) nogil:

        with nogil:
            for i in range(len(h)):
                out[i] = h3lib.h3ToParent(h[i], res)
                                        ^
    ------------------------------------------------------------

    /Users/kyle/github/mapping/h3-py/src/h3/_cy/unstable_vect.pyx:71:37: Calling gil-requiring function not allowed without gil
    gmake[2]: *** [src/h3/_cy/CMakeFiles/unstable_vect.dir/build.make:82: src/h3/_cy/unstable_vect.c] Error 1
    gmake[2]: *** Deleting file 'src/h3/_cy/unstable_vect.c'
    gmake[1]: *** [CMakeFiles/Makefile2:437: src/h3/_cy/CMakeFiles/unstable_vect.dir/all] Error 2
    gmake: *** [Makefile:171: all] Error 2
      File "/private/var/folders/6v/4_10tk8549bc7yrbycgd2c540000gn/T/pip-build-env-1vfe0biv/overlay/lib/python3.7/site-packages/skbuild/setuptools_wrap.py", line 589, in setup
        cmkr.make(make_args, env=env)
      File "/private/var/folders/6v/4_10tk8549bc7yrbycgd2c540000gn/T/pip-build-env-1vfe0biv/overlay/lib/python3.7/site-packages/skbuild/cmaker.py", line 507, in make
        os.path.abspath(CMAKE_BUILD_DIR())))
    An error occurred while building with CMake.
      Command:
        cmake --build . --target install --config Release --
      Source directory:
        /Users/kyle/github/mapping/h3-py
      Working directory:
        /Users/kyle/github/mapping/h3-py/_skbuild/macosx-10.15-x86_64-3.7/cmake-build
    Please see CMake's output for more information.
```

kylebarron: This can be fixed by adding `nogil` to `h3ToParent` in `h3lib.pxd`.

dfellis (approved):
No idea why it is failing to submit the coverage report.

src/h3/_cy/unstable_vect.pyx
ajfriend: Yeah, that's the right fix!

Adding `nogil` to function declarations just tells Cython that those functions **can** be used without the GIL, but it won't automatically release the GIL when they're called.

That's what you're doing with

```python
cpdef void h3_to_parent_vect(
    const H3Index[:] h,
    int res,
    H3Index[:] out
) nogil:
```
and when you're adding `nogil` to `h3ToParent` in `h3lib.pxd`; just saying these functions **can** be used without having the GIL.

The `with nogil` block is different. It is actually telling Cython to release and re-acquire the GIL before and after that block. That's what your error is about above: you're releasing the GIL, but as far as Cython knows (before you added `nogil`), `h3ToParent` requires the GIL.

ajfriend:
And one open question is what we do with invalid inputs for these vectorized functions. For the non-vectorized ones, we might raise an exception.

Errors might be:
- the input array contains an invalid h3 cell
- an input cell already has lower resolution than the given resolution.

For the vectorized ones, do we just return `H3_NULL`? Fail the whole function call?

ajfriend:
Another question: Would we ever want to vectorize the resolution parameter in `3_to_parent`, or are we OK with forcing it to always be a single `int`?

src/h3/_cy/unstable_vect.pyx
kylebarron: Yeah that makes a ton of sense because presumably there will be functions that partially interact with Python and that can only release the GIL for portions of the function.

isaacbrodsky (approved):

kylebarron:
Re exceptions, I think it would be fine to fail the whole function call, but open to suggestions. It might be good to look into how numpy handles exceptions.

As it is in this PR, it appears that invalid values are coerced to 0?
```py
vect.h3_to_parent(np.array([17584386400583679], dtype=np.uint64), 8)
# array([0], dtype=uint64)
```

Input arrays that aren't `np.uint64` will get a `ValueError`:
```
vect.h3_to_parent(np.array([1], dtype=np.uint8), 8)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-71-3751d4de5ad0> in <module>
----> 1 vect.h3_to_parent(np.array([1], dtype=np.uint8), 8)

~/local/anaconda3/lib/python3.7/site-packages/h3/unstable/vect.py in h3_to_parent(h, res)
     21     out = np.zeros(len(h), dtype='uint64')
     22 
---> 23     _vect.h3_to_parent_vect(h, res, out)
     24 
     25     return out

unstable_vect.pyx in h3._cy.unstable_vect.h3_to_parent_vect()

ValueError: Buffer dtype mismatch, expected 'const H3Index' but got 'unsigned char'
```


> Would we ever want to vectorize the resolution parameter in `h3_to_parent`, or are we OK with forcing it to always be a single `int`?

I think it's better to keep it an `int` for simplicity. Someone might care about a few resolution levels, but the number of hex values will dwarf the number of resolution values, so IMO better to vectorize over hex and leave resolution as `int`.

kylebarron:
To sum up a bit more research: it's possible to iterate over arbitrary multi-dimensional arrays with the C-API. Example on my [`vectorize-ufunc`](https://github.com/kylebarron/h3-py/blob/075df7ada80128e23096ba84fccdc1d4ed52d3e2/src/h3/_cy/unstable_vect.pyx#L72-L94) branch.

It should also be possible [using `np.nditer`](https://stackoverflow.com/a/39058906), but that may be slower than the C-API.

Possibly the simplest approach, if the variety of inputs is less complex but you still want to accept arbitrary dimension inputs, you can use `np.ravel` to get a 1-dimensional view on n-dimensional input, and then `reshape` back to the original input. https://stackoverflow.com/a/33316468

Since `np.ravel` can return data in the same order, a function like `geoToH3` could take two arbitrary-dimension arrays for `lat` and `lon`, check that the shapes are the same, and then use `np.ravel` to iterate over the data and return an `H3Index` array of the same shape as the input `lat` and `lon`.

All of the above approaches require a dependency on Numpy at build time and runtime, so presumably there would be a flag from Python that only runs this code if the user has Numpy installed.

kylebarron:
I was looking at this PR again and realized that it wasn't as far as I thought from merge-ready, so I figured I try to push to get it over the finish line.

**Final changes:**

- Add a default argument for `h3_to_parent`. If `res` isn't provided, then the parent one resolution up _from each cell_ is used. Note that this happens per-cell, so if you have an array of cells at resolution `[9, 8, 7]`, and `res` is `None`, then the results will be at resolution `[8, 7, 6]`.
- Allow `res` to `h3_to_parent` to be an array. This allows `res` to vary for each hex. I haven't tested broadcasting where the hex input and `res` are different shapes (I expect it would crash).
- Add a test file

Let me know if you have any more comments.

I also did the original benchmarks again and found similar results (except it looks like Python 3.8 has some `list` improvements 😄 )

```py
import h3.api.numpy_int as h3
import h3.unstable.vect as vect
import numpy as np

N = 100_000
K = 10

lats, longs = np.random.uniform(0, 90, N), np.random.uniform(0, 90, N)
h3_index = vect.geo_to_h3(lats, longs, 9)

%%timeit
out_loop = []
for i in range(len(h3_index)):
    out_loop.append(h3.h3_to_parent(h3_index[i], 8))
# 55.5 ms ± 1.66 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

vec_fn = np.vectorize(h3.h3_to_parent)
%timeit vec_fn(h3_index, 8)
# 41.2 ms ± 960 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit out_vect = vect.h3_to_parent(h3_index, 8)
# 411 µs ± 16.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

np.array_equal(out_loop, out_vect)
# True
```

kylebarron:
One more note: it might be nice to coerce input in these functions to a numpy array? There are some numpy-like types (like some pandas types) that I've found don't work automatically with the current implementation, so a simple `np.array()` on the input could make sense. That would also make it work with arbitrary Python lists.

ajfriend:
This is great! I'm taking a look now.

> One more note: it might be nice to coerce input in these functions to a numpy array? There are some numpy-like types (like some pandas types) that I've found don't work automatically with the current implementation, so a simple `np.array()` on the input could make sense. That would also make it work with arbitrary Python lists.

Wrapping with `np.array()` makes sense. Do you have a few examples of it failing with pandas types? We could make an issue with these (and the `list` example) to make sure we're handling these cases.

Also, just thinking about long term, I think we'd ultimately like to have the vectorized API use memoryviews so that we don't have a strict dependency on `numpy`. Using `numpy` right now totally makes sense as we're playing around with things, I just want to call out that at some point, we'll want to try to see if we can get memoryviews working.

src/h3/_cy/unstable_vect.pyx
```diff
@@ -57,3 +58,32 @@ cpdef void geo_to_h3_vect(
         for i in range(len(lat)):
             c = deg2coord(lat[i], lng[i])
             out[i] = h3lib.geoToH3(&c, res)
+
+
+@boundscheck(False)
+@wraparound(False)
+cpdef void h3_to_parent_vect(
+    const H3int[:] h,
+    unsigned char[:] res,
```
ajfriend, line 67: Can we just use `int` here, to match with the H3 C API?

src/h3/_cy/unstable_vect.pyx
```diff
@@ -57,3 +58,32 @@ cpdef void geo_to_h3_vect(
...
+    with nogil:
+        for i in range(len(h)):
+            out[i] = h3lib.h3ToParent(h[i], res[i])
+
+
+@boundscheck(False)
+@wraparound(False)
+cpdef void h3_get_resolution_vect(
+    const H3int[:] h,
+    unsigned char[:] out,
```
ajfriend, line 82: Can we just use `int` here, to match with the H3 C API?

src/h3/unstable/vect.py
```diff
@@ -3,9 +3,57 @@
...
+    -------
+    array of H3Cells
+    """
+    h = np.array(h, dtype=np.uint64)
+    out = np.zeros(len(h), dtype=np.uint64)
+
+    if res is None:
+        res = h3_get_resolution(h) - 1
+    else:
+        res = np.full(h.shape, res, dtype=np.uint8)
```
ajfriend, line 27: Can we just use `int` here, to match with the H3 C API?

src/h3/unstable/vect.py
```diff
@@ -3,9 +3,57 @@
...
+    Parameters
+    ----------
+    h : H3Cell
+
+    Returns
+    -------
+    array of uint8
+    """
+    h = np.array(h, dtype=np.uint64)
+    out = np.zeros(len(h), dtype=np.uint8)
```
ajfriend, line 47: Can we just use `int` here, to match with the H3 C API?

tests/test_unstable_vect.py
```diff
@@ -0,0 +1,63 @@
...
+    arr2 = np.array(list(map(h3.h3_to_parent, h)), dtype=np.uint64)
+    assert np.array_equal(arr1, arr2)
+
+    # Test with a number passed to res
+    arr1 = h3_vect.h3_to_parent(h, 7)
+    arr2 = np.array([h3.h3_to_parent(c, 7) for c in h], dtype=np.uint64)
+    assert np.array_equal(arr1, arr2)
+
+    # Test with an array passed to res
+    arr1 = h3_vect.h3_to_parent(h, np.array([7, 5], np.uint8))
```
ajfriend, line 26: Can we just use `int` here, to match with the H3 C API?

tests/test_unstable_vect.py
```diff
@@ -0,0 +1,63 @@
...
+    # Parent cells are 8, 7
+    parents = h3_vect.h3_to_parent(h)
+    assert list(h3_vect.h3_get_resolution(parents)) == [8, 7]
+
+
+def test_h3_get_resolution():
+    h = np.array([617700169958555647], np.uint64)
+
+    arr1 = h3_vect.h3_get_resolution(h)
+    arr2 = np.array(list(map(h3.h3_get_resolution, h)), dtype=np.uint8)
```
ajfriend, line 62: Can we just use `int` here, to match with the H3 C API?

ajfriend (changes requested):
Looks good! Let's just use `int` instead of `unsigned char`, as that matches the C lib.

ajfriend:
Also, `Py_ssize_t` is a nice catch! I've got to make a note to use that elsewhere in the library.

ajfriend:
I ran your benchmarks, adding a list comprehension (since they're typically faster than loops), and `np.fromiter`. The list comprehension is a bit faster, but the vectorized version is still the obvious winner.

```python
import h3.api.numpy_int as h3
import h3.unstable.vect as vect
import numpy as np

N = 100_000

lats, longs = np.random.uniform(0, 90, N), np.random.uniform(0, 90, N)
h3_index = vect.geo_to_h3(lats, longs, 9)

vec_fn = np.vectorize(h3.h3_to_parent)

%%timeit
out_loop = []
for i in range(len(h3_index)):
    out_loop.append(h3.h3_to_parent(h3_index[i], 8))
# 101 ms ± 3.2 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit
out_list_comp = [h3.h3_to_parent(h, 8) for h in h3_index]
# 88.4 ms ± 4.31 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit
out_fromiter = np.fromiter((h3.h3_to_parent(h, 8) for h in h3_index), np.uint64)
# 90.8 ms ± 2.64 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit
out_vec_fn = vec_fn(h3_index, 8)
# 76.8 ms ± 3.97 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit
out_vect = vect.h3_to_parent(h3_index, 8)
# 557 µs ± 26.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

kylebarron:
> Looks good! Let's just use `int` instead of `unsigned char`, as that matches the C lib.

Ok. I figured for `res` specifically `unsigned char` might be ideal because `res` will never be >255.

> Wrapping with `np.array()` makes sense. Do you have a few examples of it failing with pandas types? We could make an issue with these (and the `list` example) to make sure we're handling these cases.

The current revision of the code coerces with `np.array`, i.e. [here](https://github.com/uber/h3-py/pull/166/files#diff-54cf43b22fe77253ccce42713b84322d62eb1e2e2154885581e7441be5dfe7d4R21), but if you take that out, I believe passing any Pandas Series, i.e. just `pd.Series(np.array(data))` it'll fail. I think the issue is that the _Numpy array_ underlying the series implements the buffer protocol, but the Series itself doesn't. So the "right" way there would be to pass the underlying array using `series.values`, or you can coerce the input to a numpy array.

I added a test case with list input, but not with pandas because I didn't want to add a (test) dependency.

> Also, just thinking about long term, I think we'd ultimately like to have the vectorized API use memoryviews so that we don't have a strict dependency on `numpy`. Using `numpy` right now totally makes sense as we're playing around with things, I just want to call out that at some point, we'll want to try to see if we can get memoryviews working.

Well right now the Cython code uses memoryviews directly, i.e. there's still no build-time dependency on Numpy. Numpy is still used at runtime on the Python side, but in my opinion I don't see appeal of removing numpy there. Is there anyone who cares to use the vectorized bindings and isn't using Numpy? Numpy is such a fundamental part of numeric computing in Python that I don't really see the use case. That said, it does make sense to me that the non-vectorized API should still make Numpy optional.

> Also, `Py_ssize_t` is a nice catch! I've got to make a note to use that elsewhere in the library.

I just remember seeing some Cython docs somewhere saying that's the right type to use for a loop index. 🤷‍♂️ 

> The list comprehension is a bit faster, but the vectorized version is still the obvious winner.

The benchmarks are really striking to me; it goes to show that when you have really fast C code, you can really easily bog it down with Python object overhead.

kylebarron:
> Also, just thinking about long term, I think we'd ultimately like to have the vectorized API use memoryviews so that we don't have a strict dependency on `numpy`

Also doesn't this imply that every time someone passes a numpy array to the vectorized API, they'd have to call `np.asarray` on the output to use it with numpy? Either that or manually create an `out` numpy array of the right type to pass to the vectorized API to be filled?

ajfriend:
> Ok. I figured for `res` specifically `unsigned char` might be ideal because `res` will never be >255.

I've gone back and forth on this myself. :) The general advice seems to be to avoid `unsigned` types if you'll ever be doing arithmetic or comparisons, since the `unsigned` can give unexpected behavior.

> Well right now the Cython code uses memoryviews directly, i.e. there's still no build-time dependency on Numpy. Numpy is still used at runtime on the Python side, but in my opinion I don't see appeal of removing numpy there. Is there anyone who cares to use the vectorized bindings and isn't using Numpy? Numpy is such a fundamental part of numeric computing in Python that I don't really see the use case. That said, it does make sense to me that the non-vectorized API should still make Numpy optional.

Agreed that `numpy` is pretty standard, but I also think it's possible to provide a vectorized memoryview-only interface that doesn't have the dependency, and then a `numpy` interface (which outputs numpy arrays and so wouldn't require a call to`np.asarray`) would just be a simple wrapper around the memoryview interface. This is basically what I do with the current (unvectorized) `numpy` interface. If that still works and doesn't have any major downsides, I think it's worth doing even if the potential audience is pretty small.

ajfriend (approved):

ajfriend:
Hmmm... I forget how we wanted to handle these "unstable API" changes. I guess I'll do a follow-up PR and release a `v3.7.2`.
Sound good, @kylebarron @dfellis @isaacbrodsky @nrabinowitz ?

kylebarron:
> Agreed that `numpy` is pretty standard, but I also think it's possible to provide a vectorized memoryview-only interface that doesn't have the dependency

I don't have any objections to this; I think you just strive further than I would to make these bindings as unopinionated as possible 😄

PR #167, opened 2020-09-04
Vectorization attempt v2

kylebarron:
This PR is meant to be a more general PR than #166. 

- Use `np.ravel` then later `array.reshape` (from the Python side) to support any dimensionality of input data
- Multi-dimensional example of `hex_ring`. Flatten input array to 1D, then create `out` as a 2D array, where the second dimension holds the values of `hex_ring` for each input, then reshape back to original shape + 1 extra dimension in Python side.
- I also tested using Numpy syntax withing Cython, but that needs Numpy headers, thus being harder to compile, and I was surprised to learn that using memoryviews is actually preferable anyways. See: https://stackoverflow.com/a/49804276 and https://stackoverflow.com/a/37497998.
- The vectorization code in this PR wraps the internal Cython API instead of the `h3lib` API directly. This has the benefit of simplicity, as the low-level bindings to `h3lib` don't need to be repeated, but _appears_ to require the GIL. It looks like at least `create_ptr` requires the GIL, so `_cy.cells.ring` can't be used in a loop within `with nogil`. The better solution here is probably to release the GIL within `_cy.cells.ring` directly when it calls `h3lib.hexRing`, ref https://github.com/uber/h3-py/issues/140

Todo:

- [x] Use `cimport ring`; requires a pxd file for `cells.pyx`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,8 +19,3 @@ add_subdirectory(src/h3lib)
 # Build the rest (other than the core library dependency) as shared
 set(BUILD_SHARED_LIBS ON)
 add_subdirectory(src/h3)
-
-# Include Numpy
-# TODO: make optional
-find_package(NumPy REQUIRED)
-target_include_directories(h3 PUBLIC ${NumPy_INCLUDE_DIRS})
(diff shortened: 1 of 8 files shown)


kylebarron:
I'm going to close this PR. I don't think supporting higher dimensional (more than 1D) array input is necessary. And I'd rather just focus on #166.

PR #171, opened 2020-09-27
New cell area functions with h3==3.7

ajfriend:


src/h3/_cy/h3lib.pxd
```diff
@@ -149,13 +141,28 @@ cdef extern from "h3api.h":
...
     int maxFaceCount(H3Index h3)
-
     void h3GetFaces(H3Index h3, int *out)
 
     int experimentalH3ToLocalIj(H3Index origin, H3Index h3, CoordIJ *out)
-
     int experimentalLocalIjToH3(H3Index origin, const CoordIJ *ij, H3Index *out)
+
+    double hexAreaKm2(int res) nogil
+    double hexAreaM2(int res) nogil
```
isaacbrodsky, line 153: It occurs to me after adding this in Java that the existing hexArea functions don't have a radians or radians^2 unit option in the C library.

isaacbrodsky (approved):

src/h3/_cy/h3lib.pxd
ajfriend: That's true, but I figured they would be useful for folks if they were looking at, say, other planets with a different radius.

Should we add a `hexAreaRads2`?

PR #183, opened 2021-03-24
Add linux aarch64 wheel support

odidev:
Added linux aarch64 wheel support.
Related to #182. @ajfriend Could you please review this PR?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,3 +46,46 @@ jobs:
 
       - name: Tests
         run: pytest --cov=h3 --full-trace
+
+  test-arm64:
+    name: "Test ${{ matrix.pyver }} aarch64"
+    strategy:
+      matrix:
+        pyver: [cp36-cp36m, cp37-cp37m, cp38-cp38, cp39-cp39]
+      fail-fast: false
+    runs-on: ubuntu-latest
+    env:
+      py: /opt/python/${{ matrix.pyver }}/bin/python
+      img: quay.io/pypa/manylinux2014_aarch64
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Checkout submodules
+      run: |
+        git submodule init
+        git submodule update
+    - name: Set up QEMU
+      id: qemu
+      uses: docker/setup-qemu-action@v1
+    - name: Install tools
+      run: |
+        docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+          ${{ env.img }} ${{ env.py }} -m pip install -U  pip build
+    - name: Install dependencies
+      run: |
+            docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+            ${{ env.img }} \
+            bash -exc '${{ env.py }} -m pip install virtualenv && ${{ env.py }} -m venv .env && \
+            source .env/bin/activate && \
+            python -m pip install --upgrade pip && \
+            deactivate'
+    - name: Test with pytest
+      run: |
+          docker run --rm -v ${{ github.workspace }}:/ws:rw --workdir=/ws \
+            ${{ env.img }} \
+            bash -exc '\
+            source .env/bin/activate && \
+            pip install --upgrade pip setuptools wheel && \
+            pip install .[all] && \
+            pytest --cov=h3 --full-trace && \
+            deactivate'
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -91,3 +91,55 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.pypi_password }}
         run: |
           twine upload wheelhouse/*
+
+  arm64_make_wheels:
+    name: build aarch64 wheel
+    runs-on: [ubuntu-latest]
+    strategy:
+      matrix:
+       python_tag: [ "cp36-*", "cp37-*", "cp38-*", "cp39-*"]
+
+    steps:
+
+      ## Setup Env
+
+      - uses: actions/checkout@v2.3.3
+        with:
+          submodules: recursive
+
+      - uses: actions/setup-python@v2
+        with:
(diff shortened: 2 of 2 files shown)


ajfriend:
This is great! Thank you!

My first few thoughts:

- Would it be possible to add tests to `tests.yml`? I see that the tests run during the `wheels.yml`, but it would also be nice to have a set of tests that apply to `aarch64` but don't require building wheels.
    - I suppose `QEMU` is needed for `cibuildwheel` to do its job? Maybe the setup is similar for running non-wheel tests?
- Any reason we can't build `cp35-*`? It looks like [`cibuildwheel` supports it.](https://cibuildwheel.readthedocs.io/en/stable/#what-does-it-do)
- Any reason we need the `arm64_make_wheels` job and can't use the `make_wheels` job? It would be nice, if possible, to reuse the `## Setup Env` and `## Upload` steps, and just add to the `## Build across platforms` step. Unless you've spotted a reason why we can't do that, I could potentially help with figuring that out.
- Motivated mostly by my own laziness and ignorance, do you have any idea how easily this would extend to "macOS Apple Silicon" for `cp39-*`? If non-trivial, we can make a ticket for it; I'm sure that request is imminent :)

odidev:
Please find quoted replies: 
> * Would it be possible to add tests to `tests.yml`? I see that the tests run during the `wheels.yml`, but it would also be nice to have a set of tests that apply to `aarch64` but don't require building wheels.
>   
Done. 
>   * I suppose `QEMU` is needed for `cibuildwheel` to do its job? Maybe the setup is similar for running non-wheel tests?
>
Yes, for aarch64 qemu is required for emulation.
> * Any reason we can't build `cp35-*`? It looks like [`cibuildwheel` supports it.](https://cibuildwheel.readthedocs.io/en/stable/#what-does-it-do)
>
Python3.5 is now depreciated that is why I removed it.
> * Any reason we need the `arm64_make_wheels` job and can't use the `make_wheels` job? It would be nice, if possible, to reuse the `## Setup Env` and `## Upload` steps, and just add to the `## Build across platforms` step. Unless you've spotted a reason why we can't do that, I could potentially help with figuring that out.
>
In emulation environment the aarch64 build is slower. To make it faster I have split the job as per python version.
If we use the same job it will become slower.

> * Motivated mostly by my own laziness and ignorance, do you have any idea how easily this would extend to "macOS Apple Silicon" for `cp39-*`? If non-trivial, we can make a ticket for it; I'm sure that request is imminent :)
>
I will check and update you regarding this.

ajfriend:
Thanks! This looks basically ready to merge.

Just to check before merging, have you considered https://github.com/uraimo/run-on-arch-action ? I wonder if it could help simplify, for example, the code in `tests.yml`.

I don't think its a blocker. I'm just trying to better understand the options for the sake of future maintenance. 😄

ajfriend:
Just for future reference, I want to note these issues, which seem to have a lot of relevant information and link to a few projects that have also done this with `cibuildwheel`:

- https://github.com/joerick/cibuildwheel/pull/482
- https://github.com/joerick/cibuildwheel/issues/473

odidev:
@ajfriend, could you please review this PR? Thanks.

ajfriend:
I should be able to take a look at it this week. I definitely appreciate the effort and help on this, @odidev!

isaacbrodsky (approved):

odidev:
@ajfriend, I could see the changes are approved. Could you please let me know when this will get merged? Thanks in advance.

ajfriend:
Hey @odidev, sorry for the delay.

- Could we actually remove the changes to `tests.yml` for now? Thank you for putting that together, but that's more code than I'd want to add just to add the pre-wheel tests. I'd love to add such tests if we could find a way with a smaller footprint.
- Could you update in light of #188?
- And what would the diff look like if you did it through the existing `make_wheels` job? I know you mentioned it is slower, but if it is much simpler than having an additional `arm64_make_wheels`, I might prefer that. What's the time difference between the two options?

odidev:
> Hey @odidev, sorry for the delay.
> 
> * Could we actually remove the changes to `tests.yml` for now? Thank you for putting that together, but that's more code than I'd want to add just to add the pre-wheel tests. I'd love to add such tests if we could find a way with a smaller footprint.

Sure I can remove test.yml changes.
> * Could you update in light of #188?

The changes can be done as #188.
> * And what would the diff look like if you did it through the existing `make_wheels` job? I know you mentioned it is slower, but if it is much simpler than having an additional `arm64_make_wheels`, I might prefer that. What's the time difference between the two options?

I can surely add aarch64 jobs in `make_wheels` job, the time to build aarch64 wheels is around 1 hr. On the other hand, by adding `arm64_make_wheels` jobs and dividing aarch64 jobs for each python version, each job takes around 11 min. 
However for reducing time and to go ahead with one job for both architectures (auto and aarch64), the jobs can be divided as per python version.

@ajfriend, Please let me know your views on this and I will do the changes accordingly.

ajfriend:
> I can surely add aarch64 jobs in `make_wheels` job, the time to build aarch64 wheels is around 1 hr. On the other hand, by adding `arm64_make_wheels` jobs and dividing aarch64 jobs for each python version, each job takes around 11 min.
> However for reducing time and to go ahead with one job for both architectures (auto and aarch64), the jobs can be divided as per python version.

Ah, I see. You're splitting up the jobs by Python version. I didn't catch that that was the trick before. That's a nice idea!

It looks like #189 is working with the ~1 hour build time. Let's land #189 that for now, and potentially follow up--either in this PR or another--with something to split up the hour by Python version. I can definitely imagine that hour turnaround being annoying next time we do a release. :)

ajfriend:
It looks like we have things working in #189. Thank you for kicking off this workstream, @odidev!

After we look into #191 I'll make a release to publish the aarch64 wheels to PyPI. @odidev, could you let us know if those work for you?

Closing this. Feel free to re-open if there are any dangling issues I've fogotten.

PR #188, opened 2021-05-18
ci: cleanup and pin cibuildwheel

henryiii:
Cleaning up the file just a bit, mostly pinning cibuildwheel, as version 2.0 will drop Python 2/3.5, which might be a surprise if you just always install the latest version. It looks like you might use dependabot + pinning in actions, so you might like the action form of cibuildwheel better, ~~but it's about to move~~ from joerick/cibuildwheel to pypa/cibuildwheel, so that could be done after the move to pick up the shiny new name. :) Edit: moved, so using action.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -40,9 +40,7 @@ jobs:
           
       ## Build
 
-      - name: Build
-        run: |
-          pipx run --spec cibuildwheel==1.11.0 cibuildwheel --output-dir wheelhouse
+      - uses: pypa/cibuildwheel@v1.11.0
 
       - name: Make sdist
         if: startsWith(matrix.os, 'ubuntu')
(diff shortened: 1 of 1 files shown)


dfellis:
So this refactoring doesn't seem to work (yet).

henryiii:
I thought pipx recognized a version number without having to use --spec. The action doesn't have to worry about that.

dfellis (approved):

isaacbrodsky (approved):

ajfriend:
Awesome! Thanks @henryiii! Would you happen to know if it's similarly easy to extend this to also build for `aarch64` and Apple Silicon?

ajfriend (approved):

ajfriend:
But I should also figure out how this overlaps with #183

henryiii:
It should be pretty easy, depends on if you support Aarch64 (#183) and if there are any dependencies that don't support Apple Silicon yet. I could draft an example PR, possibly.

PR #189, opened 2021-06-08
ci: add ARM builds

### LINKED ISSUE #182: Add support to release aarch64 wheels
Problem
---------
On aarch64, pip install h3-py builds the wheels from source code and then install it. It requires user to have development environment installed on his system. also, it take some time to build the wheels than downloading and extracting the wheels from pypi.

Resolution
-----------
On aarch64, pip install h3-py should download the wheels from pypi

@ajfriend  , please let me know your interest on releasing aarch64 wheels. I can help in this.

henryiii:
Rough idea based on https://github.com/uber/h3-py/pull/188#issuecomment-856365308

Matrix element names will change a bit.

Closes #182.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -6,27 +6,58 @@ on:
   pull_request:
     branches: [master]
   release:
+    types:
+      - published
 
-env:
-  CIBW_TEST_REQUIRES: pytest numpy
-  CIBW_TEST_COMMAND: pytest {project}/tests
-  CIBW_SKIP: pp* cp27*-win* *-win32 *-manylinux_i686
 
 jobs:
+  make_sdist:
+    name: SDist
+    runs-on: ${{ matrix.os }}
+
+    steps:
+
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Make sdist
+        run: |
+          pipx run build --sdist
+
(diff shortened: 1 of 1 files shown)


henryiii:
Oops, forgot QEMU.

henryiii:
Looks like we should probably work on scikit-build's Apple Silicon support (see https://github.com/pypa/cibuildwheel/issues/582 ). scikit-build/cmake-python-distributions was adapted to AS, now waiting on upstream Kitware for scikit-build/ninja-python-distributions, though we probably could move forward with scikit-build/scikit-build while waiting on that. See https://github.com/scikit-build/scikit-build/pull/530 .

The Aarch64 build should be fine.

ajfriend:
Awesome. Thanks!

And thanks for the pointers, I'll make a ticket to check back in to follow up on AS.

Let's land this for now (should be good to go as is, yeah?), and follow up with @odidev and #183 to explore if it makes sense to split out the Python versions on `aarch64` to speed up the CI run time.

henryiii:
I've attempted a refactor to split the Arm job up more, along with nicer names. This also will avoid uploading any wheels at all if one of the jobs breaks (which didn't happen before).

isaacbrodsky (approved):

ajfriend:
Awesome. This is great! Thanks a lot for putting this together @henryiii!

PR #197, opened 2021-08-17
Getting docs to a good starting point

ajfriend:
Here's a starting point for the docs (builds without failures due to incorrectly formatted docstrings).

Definitely more work to do. I'm not sure if we should just land this now and iterate on the docs (potentially in an async fashion), or if we should continue to polish up this PR.

You can see the docs built from this PR at: https://ajfriend.github.io/h3-py (I've been using this as sort of a docs staging area)

An older, very incomplete version is currently live at https://uber.github.io/h3-py

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -3,10 +3,22 @@
 # See the documentation for help and more options:
 # https://jupyterbook.org/customize/config.html
 
-title: h3-py  # The title of the book. Will be placed in the left navbar.
+title: h3-py
 author: 'Uber Technologies'
-copyright: "2021"  # Copyright year to be placed in the footer
-logo: logo.svg  # A path to the book logo
+copyright: "2021"
+logo: logo.svg
+
+html:
+  favicon: logo.svg
+  home_page_in_navbar: false
+  use_edit_page_button: true
+  use_repository_button: true
+  baseurl: https://uber.github.io/h3-py/
+
+repository:
+  url: https://github.com/uber/h3-py
+  branch: master
+  path_to_book: docs
 
 sphinx:
   extra_extensions:
@@ -14,4 +26,3 @@ sphinx:
     - 'sphinx.ext.autosummary'
     - 'sphinx.ext.napoleon'
     - 'sphinx.ext.viewcode'
-
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -14,3 +14,4 @@ sphinx:
     - 'sphinx.ext.autosummary'
     - 'sphinx.ext.napoleon'
     - 'sphinx.ext.viewcode'
+
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -33,7 +33,9 @@ jobs:
       # Build the book
       - name: Build the book
         run: |
-          jupyter-book build docs/ --warningiserror --keep-going --all
+          jupyter-book build docs/
+
+      # jupyter-book build docs/ --warningiserror --keep-going --all
 
       # Deploy the book's HTML to gh-pages branch
       - name: GitHub Pages action
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -33,7 +33,7 @@ jobs:
       # Build the book
       - name: Build the book
         run: |
-          jupyter-book build docs/
+          jupyter-book build docs/ --warningiserror --keep-going --all
 
       # Deploy the book's HTML to gh-pages branch
       - name: GitHub Pages action
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - master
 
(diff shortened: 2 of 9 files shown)


ajfriend:
Also, @kylebarron, maybe there's some overlap with your #194 in terms of making sure the docstrings are properly formatted and match up with the type stubs.

ajfriend:
Also, there is some nice functionality with JupyerBook that lets you launch notebook examples directly into a live cloud service, which would be awesome for the examples repo.

For example, check out the "Launch Notebook" button on this page: https://python.quantecon.org/geom_series.html

ajfriend:
Another idea: Juypter Book lets you output to multiple formats. It might be nice to publish a PDF version of the docs with each `h3-py` release. Definitely nice for historical documentation purposes.

docs/_config.yml
```diff
@@ -3,10 +3,22 @@
...
-logo: logo.svg  # A path to the book logo
+copyright: "2021"
+logo: logo.svg
+
+html:
+  favicon: logo.svg
+  home_page_in_navbar: false
+  use_edit_page_button: true
+  use_repository_button: true
+  baseurl: https://uber.github.io/h3-py/
```
isaacbrodsky, line 16: I wonder if there's a way to include this on the h3geo.org domain too - maybe a subdomain (like py.h3geo.org?)

isaacbrodsky (approved):

docs/_config.yml
ajfriend: That'd be nice!

kylebarron (approved):
- Cool! I've never used Jupyter Book for docs before (I've only used mkdocs, [example](https://cogeotiff.github.io/rio-tiler/))
- How does docstring generation work? Is it interpreting the docstring as Markdown? Why do you need double ``` `` ``` instead of a single `` ` ``?
- Judging from `.. code-block:: text` ... this is using sphinx to generate the api docs? So it's kinda a mix of RST for docstrings, Markdown for prose?
- I'd suggest checking out [`pydocstyle`](https://www.pydocstyle.org/en/stable/usage.html) for CI if you plan to autogenerate docs from docstrings. Might make it simpler to verify docstrings are as expected (but I haven't used it myself).
- Re types: I don't really understand the intersection between RST/sphinx/jupyter book in this docs setup, but I've gotten type inference to work in Sphinx docs generation [before](https://stackoverflow.com/a/51312475) (3 years ago 😄) via https://github.com/agronholm/sphinx-autodoc-typehints. But I'm not sure if that works at all via stub files, or just using in-code type hints, and might be irrelevant if you're not using autodoc.

docs/_config.yml
```diff
@@ -3,10 +3,22 @@
...
+
+repository:
+  url: https://github.com/uber/h3-py
+  branch: master
+  path_to_book: docs
 
 sphinx:
   extra_extensions:
```
kylebarron, line 24: Oh I see you are using `autodoc`...

kylebarron:
Took a quick look; I checked out your repo and added the `sphinx_autodoc_typehints` dependency. It does work out of the box with in-line types, e.g.
```py
    def versions() -> Dict[str, str]:
```
shows 
image

But 

1. I don't have the impression that you'd be interested in in-line types, since I think this still supports 2.7.
2. It could be hard to get the generic typing to work for in-line types (though I'm not sure it works for the type stubs anyways), and the typehints plugin complains with `WARNING: Cannot treat a function defined as a local function: "h3.versions"  (use @functools.wraps`. I think this is because of the API factory function.
3. Type stubs aren't supported. https://github.com/agronholm/sphinx-autodoc-typehints/issues/161

ajfriend:
Thanks @kylebarron 

> * How does docstring generation work? Is it interpreting the docstring as Markdown? Why do you need double `` `` `` instead of a single `` ` ``?
> * Judging from `.. code-block:: text` ... this is using sphinx to generate the api docs? So it's kinda a mix of RST for docstrings, Markdown for prose?

JupyterBook uses the [MyST Parser](https://myst-parser.readthedocs.io/) so that you can use markdown instead of RST, which I'll do almost anything to avoid :)
It's a fairly complete replacement, but with one major exception being docstrings, which is why you need those `` `` `` (in the docstrings, but not in the website markdown files).


> * I'd suggest checking out [`pydocstyle`](https://www.pydocstyle.org/en/stable/usage.html) for CI if you plan to autogenerate docs from docstrings. Might make it simpler to verify docstrings are as expected (but I haven't used it myself).

I'll check it out, thanks!

> * Re types: I don't really understand the intersection between RST/sphinx/jupyter book in this docs setup, but I've gotten type inference to work in Sphinx docs generation [before](https://stackoverflow.com/a/51312475) (3 years ago 😄) via https://github.com/agronholm/sphinx-autodoc-typehints. But I'm not sure if that works at all via stub files, or just using in-code type hints, and might be irrelevant if you're not using autodoc.

Since this doesn't seem to be straightforward, let's defer the typing stuff (at least for the docs) to later.

> 1. I don't have the impression that you'd be interested in in-line types, since I think this still supports 2.7.

And I don't think we're ready to drop 2.7 just yet, as there is still a large user base. (I think)

PR #211, opened 2022-02-17
Trying 4.0

ajfriend:


ajfriend:
Trying to add a second submodule for the 4.0 C lib and building that separately didn't seem to work. :(

PR #212, opened 2022-02-17
Starting work on 4.0

ajfriend:


.github/workflows/tests.yml
```diff
@@ -14,10 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
-        exclude:
-          - os: windows-latest
-            python-version: 2.7
+        python-version: [3.7, 3.8, 3.9]
```
jmccartin, line 17: Any chance Python 3.10 can be added here, and to the list of pre-built wheels? Can raise this as a separate issue if you'd prefer.

ajfriend: Yup. I pared things back just to try to get the compilation working. I'm planning on adding back in all the wheels and other actions.

ajfriend:
Abandoning, as this was subsumed by #238

PR #213, opened 2022-02-17
Class-based `api_functions` for better tooling support

kylebarron:
### Changes

- Change `_api_functions` from a function that directly modifies `globals` to a class. This class is instantiated inside `__init__`, so any import like `from h3 import basic_int` will now import a class instead of a module, though functionally there's no difference.
- Add minimal `_public_api.py` files to bind class methods onto module scope. This solves the IDE/linter problem for function imports by having nothing added to globals at runtime. These need to be copied manually to stay in sync. A test is added to verify that each `_public_api.py` has the same text.

#### Backwards compatible

All existing tests pass.

#### Lint tooling support

Adds support for linters and IDE tooling. (I.e. would close https://github.com/uber/h3-py/issues/180)

Here are some screenshots from vscode, but it should be similar in different IDEs.

**Current intellisense** is unable to see any functions. Since everything is done at runtime, there's nothing a linter can see.

Screen Shot 2022-02-17 at 1 11 42 PM

**With this PR**, an IDE can see methods off of `basic_int`

Screen Shot 2022-02-17 at 1 29 19 PM

And also for plain function imports
image
image

Closes https://github.com/uber/h3-py/issues/80, closes https://github.com/uber/h3-py/issues/180


kylebarron:
I should also note that an easy way to make the linting/typing/tooling work with function imports as well is just to define them statically. E.g.

```py
basic_int = _API_FUNCTIONS[int, Set[int]](
    _in_scalar=_id,
    _out_scalar=_id,
    _in_collection=_in_collection,
    _out_unordered=set,  # todo: should this be an (immutable) frozenset?
    _out_ordered=list,  # todo: should this be an (immutable) tuple?
)

versions = basic_int.versions
string_to_h3 = basic_int.string_to_h3
...
```

In my view it wouldn't be hard to have a build step or CI step that adds this block to the bottom of each API file and checks that they're up to date. And it would be worth it to make the tooling work well in all cases.

ajfriend:
This looks nice!

In thinking through solutions to the IDE integration problem, would you happen to know what the requirements are from the IDE to be able to provide the nice functionality? Does it always come down to needing a static file that the IDE can read from, or is there a programmatic API where we'd just need to provide the appropriate information?

kylebarron:
> In thinking through solutions to the IDE integration problem, would you happen to know what the requirements are from the IDE to be able to provide the nice functionality?

I don't know exactly. I think in general tooling is unable to sense dynamically-generated APIs that don't have static Python imports. For example Pylint has an [option called `generated-members`](https://pylint.pycqa.org/en/latest/technical_reference/features.html#typecheck-checker-options), which allows you to specify modules like `h3` that it won't be able to find otherwise.

> Does it always come down to needing a static file that the IDE can read from, or is there a programmatic API where we'd just need to provide the appropriate information?

I think the programmatic API is the static file 😄 . 

Note that in this PR in the [first commit](https://github.com/uber/h3-py/pull/213/commits/4efd64b14ddff2c55058614a34ed4b50ad43cecc), tooling only worked for the class import (i.e. `from h3 import basic_int`) and not for any functions (i.e. `from h3 import k_ring`) because those functions were still generated at runtime. So it seems like the only way to provide information to tooling is by having static module members 🤷‍♂️ .

scripts/api_functions.py
```diff
@@ -0,0 +1,91 @@
...
+"""
+
+
+def read_api(api_template_path: Path) -> List[str]:
+    """Read public API from _api_template.py"""
+    with open(api_template_path) as f:
+        lines = f.readlines()
+
+    # Methods within _API_FUNCTIONS class
+    method_lines = [line for line in lines if line.startswith("    def")]
```
isaacbrodsky, line 21: Can this be done with parsing the file AST? Probably fine to do this way too, especially if we have tests, but AST or reflection would be great.

kylebarron: It's definitely possible to do with AST. And actually the easiest approach here might be introspection, since we can run the import. So we should be able to do
```py
from h3.api._api_functions import _API_FUNCTIONS
instance = _API_FUNCTIONS(...)
methods = [
	method for method in dir(instance) if not method.startswith('_')
]
```

kylebarron:
@ajfriend do you think this would be easier to review if we split the change to a class and the addition of types into two separate PRs?

ajfriend:
> @ajfriend do you think this would be easier to review if we split the change to a class and the addition of types into two separate PRs?

Yeah, I would prefer to do things that way.

kylebarron:
> Yeah, I would prefer to do things that way.

I removed all type hinting from this PR. This PR is still blocked on flake8 lint errors, where the generated bindings have lines that are too long. This could be solved by either:

- #215
- Ignoring flake8, but this can't be done within a block, so we'd either ignore line length errors in the entire file or need a flake8 ignore on each line.

ajfriend:
Thanks! I have one alternate approach that I'd like to consider that's module-based (instead of class-based) but hopefully gets us to the same outcome. I'll try to put up a PR soon and we can compare approaches.

kylebarron:
I think the main reasons why I went with a class-based instead of module-based approach are:

1. You can bind the API constructors (I mean like `_in_scalar`, `out_scalar`, etc) to `self` really easily, while I don't quite see how you could do this with a module approach without resorting to a dynamic approach.
2. A class approach allows you to [subclass from `Generic`](https://github.com/uber/h3-py/pull/216/files#diff-9368e938da0044df4c0a4b85f0d91c3d8be718b1258a2367ec88967580171842R64) which allows for really easy typing for all APIs. Just define the method types once based on the generic types, and then parametrize with from each API with something like `_API_FUNCTIONS[str, Set[str], List[str]]`. This is the simplest way in my opinion to add type support for each API with minimal duplication.

ajfriend:
Apologies for taking so long to get to this, but I do like your approach here. I think the `_API_FUNCTIONS` class is definitely the way to go. 

I think I agree what you're doing here is better than what I was doing in #218, even if I don't love the copying in the `_public_api.py` files. I guess there's just no perfect solution here. 😕 

Although, I do wonder if some fancy import system hacking could allow us to get rid of the `_public_api.py` files by having `basic_int` be a class (like you originally had it), but still allowing for import statements like `from h3.api.basic_int import k_ring`. (And correct me if I'm wrong, but that's the whole reason you have `_public_api.py` here, right?) But we don't need to get into that here, I'm happy to land this PR as is.

Thanks again for all the work on this! Let me know if you have anything else you'd like to do on this, otherwise I can land in the next day or so.

src/h3/api/numpy_int/_binding.py
```diff
@@ -19,25 +19,29 @@
...
     # array is copied only if dtype does not match
     # `list`s should work, but not `set`s of integers
     return np.asarray(x, dtype='uint64')
 
 
-_api_functions(
+_binding = _API_FUNCTIONS(
     _in_scalar = _id,
-    _out_scalar = _id,
+    _out_scalar = _to_uint64_scalar,
```
ajfriend, line 43: What was your use case for this? Personally, I wasn't sure if it was better to have the scalars as Python `int`s or `numpy.uint64`s, so I just defaulted to having the scalars match the `basic_int` API.

But I'm curious to know if you have an example where this works better. Maybe we can even turn it into a test.

ajfriend (approved):

ajfriend:
And @kylebarron, any idea if there's a way to have tests for this IDE and linter support?

dfellis:
> Apologies for taking so long to get to this, but I do like your approach here. I think the `_API_FUNCTIONS` class is definitely the way to go.
> 
> I think I agree what you're doing here is better than what I was doing in #218, even if I don't love the copying in the `_public_api.py` files. I guess there's just no perfect solution here. 😕
> 
> Although, I do wonder if some fancy import system hacking could allow us to get rid of the `_public_api.py` files by having `basic_int` be a class (like you originally had it), but still allowing for import statements like `from h3.api.basic_int import k_ring`. (And correct me if I'm wrong, but that's the whole reason you have `_public_api.py` here, right?) But we don't need to get into that here, I'm happy to land this PR as is.
> 
> Thanks again for all the work on this! Let me know if you have anything else you'd like to do on this, otherwise I can land in the next day or so.

If those `_public_api.py` files are actually identical, you could define once and symlink to the rest?

src/h3/api/numpy_int/_binding.py
kylebarron: Oh I think I wanted to remove this from this PR and move it to the typing PR.  We can discuss more there.

tests/test_api_bindings_match.py
```diff
@@ -0,0 +1,16 @@
...
+
+    apis = [
+        h3.api.basic_int._public_api,
+        h3.api.basic_str._public_api,
+        h3.api.memview_int._public_api,
+        h3.api.numpy_int._public_api,
+    ]
+
+    api_set = {inspect.getsource(api) for api in apis}
+    assert len(api_set) == 1
```
kylebarron, line 16: If we're symlinking the APIs, I'm not sure we need this test, but I suppose it doesn't hurt 🤷‍♂️

kylebarron:
> If those `_public_api.py` files are actually identical, you could define once and symlink to the rest?

I didn't know that symlinks would work in a Python package, but it makes sense that since the symlink is a filesystem construct, it works out of the box when constructing the wheel. And inspecting the built wheel, there are no symlinks there, just when developing. So it seems like a nice tool to use here for developing.

```
> ./env/bin/pip wheel .
> unzip -l h3-3.7.3-cp38-cp38-macosx_11_0_x86_64.whl
Archive:  h3-3.7.3-cp38-cp38-macosx_11_0_x86_64.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
...
       35  03-29-2022 15:46   h3/api/basic_int/__init__.py
      706  03-29-2022 15:46   h3/api/basic_int/_binding.py
     2692  03-29-2022 15:46   h3/api/basic_int/_public_api.py
       35  03-29-2022 15:46   h3/api/basic_str/__init__.py
      918  03-29-2022 15:46   h3/api/basic_str/_binding.py
     2692  03-29-2022 15:46   h3/api/basic_str/_public_api.py
       35  03-29-2022 15:46   h3/api/memview_int/__init__.py
      723  03-29-2022 15:46   h3/api/memview_int/_binding.py
     2692  03-29-2022 15:46   h3/api/memview_int/_public_api.py
       35  03-29-2022 15:46   h3/api/numpy_int/__init__.py
      987  03-29-2022 15:46   h3/api/numpy_int/_binding.py
     2692  03-29-2022 15:46   h3/api/numpy_int/_public_api.py
...
```

kylebarron:
> I think I agree what you're doing here is better than what I was doing in #218, even if I don't love the copying in the `_public_api.py` files. I guess there's just no perfect solution here. 😕

I think symlinking the `_public_api.py` files makes sense and should alleviate some concerns here.

> And @kylebarron, any idea if there's a way to have tests for this IDE and linter support?

Maybe... Thoughts are:

1. Should be some way to use an AST parser like the one in the standard library or [`astroid`](https://pylint.pycqa.org/projects/astroid/en/latest/index.html) (which underpins `pylint`), but I've never really dealt with AST parsing before, so it might take some work.
2. Another possibility is to use [`pylint`](https://pylint.pycqa.org/en/latest/) directly and check for `import-error` errors

	  ```py
	  # pylint_example.py
	  from h3.api.basic_int import geo_to_h3
	  geo_to_h3(1, 2, 3)
	  ```
	  
	  Using my global env's pylint (without this PR)
	  ```
	  > pylint --disable=all --enable=import-error pylint_example.py
	  ************* Module pylint_example
	  pylint_example.py:1:0: E0401: Unable to import 'h3.api.basic_int' (import-error)
	  
	  ----------------------------------------------------------------------
	  Your code has been rated at -15.00/10 (previous run: -20.00/10, +5.00)
	  ```
	  
	  Using local pylint with this PR
	  
	  ```
	  > ./env/bin/pylint --disable=all --enable=import-error pylint_example.py
	  
	  -------------------------------------------------------------------
	  Your code has been rated at 10.00/10 (previous run: 5.00/10, +5.00)
	  ```

tests/test_api_bindings_match.py
kylebarron: I wasn't sure if symlinking would matter for windows builds from source... but tests seem to pass 🤷‍♂️

dfellis:
> > If those `_public_api.py` files are actually identical, you could define once and symlink to the rest?
> 
> I didn't know that symlinks would work in a Python package, but it makes sense that since the symlink is a filesystem construct, it works out of the box when constructing the wheel. And inspecting the built wheel, there are no symlinks there, just when developing. So it seems like a nice tool to use here for developing.
> 

Symlinking like that is a neat trick for C (and other super-static languages) to fake dependency injection for your code: symlink the code you want to inject test harnesses into a test directory, write the other code in there, and then build the test application with the modified include directories so the code to be tested picks up your "DI"-like code instead of the actual dependency.

kylebarron:
> 2\. Another possibility is to use [`pylint`](https://pylint.pycqa.org/en/latest/) directly and check for `import-error` errors

I added a CI test that uses `pylint` to check for `import-error` errors on test files. I think this should be sufficient to verify that imports can be followed.

Pylint in the virtualenv on this branch gives no errors:
```
> ./env/bin/pylint --disable=all --enable=import-error tests/
```

Pylint globally gives quite a few `unable to import h3` errors:
```
> pylint --disable=all --enable=import-error tests/
************* Module tests.test_numpy_int
tests/test_numpy_int.py:1:0: E0401: Unable to import 'h3.api.numpy_int' (import-error)
************* Module tests.test_h3
tests/test_h3.py:1:0: E0401: Unable to import 'pytest' (import-error)
tests/test_h3.py:2:0: E0401: Unable to import 'pytest' (import-error)
tests/test_h3.py:4:0: E0401: Unable to import 'h3' (import-error)
tests/test_h3.py:11:4: E0401: Unable to import 'h3' (import-error)
************* Module tests.test_basic_int
tests/test_basic_int.py:1:0: E0401: Unable to import 'h3.api.basic_int' (import-error)
************* Module tests.test_to_multipoly
tests/test_to_multipoly.py:1:0: E0401: Unable to import 'h3' (import-error)
************* Module tests.test_api_bindings_match
tests/test_api_bindings_match.py:5:4: E0401: Unable to import 'h3' (import-error)
tests/test_api_bindings_match.py:6:4: E0401: Unable to import 'h3.api.numpy_int' (import-error)
************* Module tests.test_memview_int
tests/test_memview_int.py:1:0: E0401: Unable to import 'h3.api.memview_int' (import-error)
************* Module tests.test_polyfill
tests/test_polyfill.py:1:0: E0401: Unable to import 'h3' (import-error)
tests/test_polyfill.py:3:0: E0401: Unable to import 'pytest' (import-error)
************* Module tests.test_cells_and_edges
tests/test_cells_and_edges.py:1:0: E0401: Unable to import 'h3' (import-error)
tests/test_cells_and_edges.py:2:0: E0401: Unable to import 'pytest' (import-error)
tests/test_cells_and_edges.py:4:0: E0401: Unable to import 'h3' (import-error)
tests/test_cells_and_edges.py:446:4: E0401: Unable to import 'packaging.version' (import-error)
************* Module tests.test_length_area
tests/test_length_area.py:1:0: E0401: Unable to import 'h3' (import-error)
tests/test_length_area.py:2:0: E0401: Unable to import 'pytest' (import-error)
tests/test_length_area.py:4:0: E0401: Unable to import 'h3' (import-error)
************* Module tests.test_basic_str
tests/test_basic_str.py:1:0: E0401: Unable to import 'h3.api.basic_str' (import-error)
************* Module tests.test_unstable_vect
tests/test_unstable_vect.py:1:0: E0401: Unable to import 'h3.api.numpy_int' (import-error)
tests/test_unstable_vect.py:3:0: E0401: Unable to import 'pytest' (import-error)
tests/test_unstable_vect.py:9:4: E0401: Unable to import 'h3.unstable.vect' (import-error)
************* Module tests.test_collection_inputs
tests/test_collection_inputs.py:2:0: E0401: Unable to import 'pytest' (import-error)
tests/test_collection_inputs.py:3:0: E0401: Unable to import 'h3.api' (import-error)
```

isaacbrodsky (approved):

tests/test_api_bindings_match.py
ajfriend: I had a hard time telling the exact status of using symlinks with pip and wheels, but I get the impression that something does break on Windows from these discussions:

- https://github.com/pypa/pip/issues/5919
- https://discuss.python.org/t/symbolic-links-in-wheels/1945/31

But it looks like this might be the case only in older versions of Windows?

ajfriend:
This is looking great!

I played around on my machine to check that other methods of installing don't break, like via the sdist, or using `python setup.py install`. Installing directly from the github branch also works:

```
pip install --no-cache-dir git+https://github.com/kylebarron/h3-py.git@kyle/class-api-functions
```

@kylebarron @dfellis, let me know if you have thoughts on the Windows issue noted above. Otherwise, I think we can land.

ajfriend:
@kylebarron, there's one last design discussion I'd like to have (and I don't think it should block landing this PR).

If symlinking does turn out to be a tool we can use, does that make the approach in #218 more attractive? The benefit I see with using modules instead of the class in this PR is that we can avoid the function copying code in `_public_api.py`. And the duplication in #218 could be "symlinked out" like it is here. But I'm not sure if the class approach is useful for the typing work you've been considering?

Just generally interested in your thoughts on the design tradeoffs in light of this new symlinking tool we have.

tests/test_api_bindings_match.py
kylebarron: I would assume there's no issues with older Windows versions _when installing prebuilt wheels_ because the wheels _don't include symlinks_. When pip/cibuildwheel creates the windows wheels, it appears to dereference the symlinks. If you look at one of the built windows wheels from CI, you see that everything in the zipfile appears to be a hard path:

```
> unzip -l h3-3.7.3-cp39-cp39-win_amd64.whl
Archive:  h3-3.7.3-cp39-cp39-win_amd64.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
...
       95  03-29-2022 17:07   h3/api/__init__.py
    25872  03-29-2022 17:07   h3/api/_api_template.py
       36  03-29-2022 17:07   h3/api/basic_int/__init__.py
      742  03-29-2022 17:07   h3/api/basic_int/_binding.py
     2758  03-29-2022 17:07   h3/api/basic_int/_public_api.py
       36  03-29-2022 17:07   h3/api/basic_str/__init__.py
      961  03-29-2022 17:07   h3/api/basic_str/_binding.py
     2758  03-29-2022 17:07   h3/api/basic_str/_public_api.py
       36  03-29-2022 17:07   h3/api/memview_int/__init__.py
      754  03-29-2022 17:07   h3/api/memview_int/_binding.py
     2758  03-29-2022 17:07   h3/api/memview_int/_public_api.py
       36  03-29-2022 17:07   h3/api/numpy_int/__init__.py
     1030  03-29-2022 17:07   h3/api/numpy_int/_binding.py
     2758  03-29-2022 17:07   h3/api/numpy_int/_public_api.py
```

So I'd assume if there were an issue in older versions of Windows, that would only happen if the user tried to build from source. And given that we distribute wheels, I don't think it's necessary to support old Windows versions (from the link you referenceed... Vista?) to build from source.

kylebarron:
> @kylebarron @dfellis, let me know if you have thoughts on the Windows issue noted above. Otherwise, I think we can land.

I don't _think_ these are relevant to us because I don't think we have symlinks in the wheels themselves.

Even the sdist doesn't include symlinks it looks like.

The dev source has a symlink:
```
> ls -l src/h3/api/basic_str
total 16
-rw-r--r--  1 kbarron  staff   35 Mar 31 15:41 __init__.py
-rw-r--r--  1 kbarron  staff  918 Mar 31 15:41 _binding.py
lrwxr-xr-x  1 kbarron  staff   27 Mar 31 15:41 _public_api.py -> ../basic_int/_public_api.py
```

But if you extract the `.tar.gz` sdist built from CI, you see it doesn't include a symlink.

```
> ls -l src/h3/api/basic_str
total 24
-rw-r--r--@ 1 kbarron  staff    35 Mar 29 09:43 __init__.py
-rw-r--r--@ 1 kbarron  staff   918 Mar 29 09:43 _binding.py
-rw-r--r--@ 1 kbarron  staff  2692 Mar 29 09:43 _public_api.py
```

ajfriend:
> My slight opinion that the class-based approach is a little more ergonomic because the symlinks are "more out of the way". I.e. `_api_template.py` is the core of the Python bindings, and having that as a single, non-copied file still lessens the total effective lines of code of the project.
> 

Fair enough. I think lean towards the module approach to avoid the triple repetition of the function names in `_public_api.py` and `_api_template.py`.

But that's honestly more of a nit than anything. Let's land it like this.

ajfriend:
As far as the symlinks go, I agree that this change probably won't affect the vast majority of users, since most folks will install via wheels. I'm a bit worried that we're breaking some contract that installing via the zipped sdist **should** work on windows, but that seems to be a small segment of users (non-wheel and maybe an older subset of windows versions).

I don't consider that a blocker for this change. I'd just want to make a note that we should keep an eye out if we are causing any installation issues, and we can decide what to do from there.

ajfriend (approved):

kylebarron:
> Fair enough. I think lean towards the module approach to avoid the triple repetition of the function names in `_public_api.py` and `_api_template.py`.

Happy to revisit in the future if we find that's a better approach

Issue #18, opened 2018-11-22
How to use pip install h3 on Windows


maxispeicher:
Hey,
I´ve tried to install the h3 package by using pip install h3 on my windows machine. I have cmake, make and gcc installed. However cmake defaults to the Visual Studio compiler and therefore doesn´t create any make files. I´ve tried my best to fix this issue but I haven´t found a way to solve it yet.
Do you know any way how to use the pip install on a windows machine?

Many Thanks

isaacbrodsky:
Thanks for reporting this. I haven't tried installing h3-py on a Windows system yet, but it looks like it should be possible by either setting https://github.com/uber/h3-py/blob/master/.install.sh to create makefiles, or invoking `cmake --build` instead of `make`.

I'm not sure on instructions for you to test that. We would probably need to add AppVeyor CI to the project to test the Windows install.

moeppe-sp:
I´m not quite sure how to invoke the build process right after this step. I´ve already tried to tried to change the .intsall.sh file to create Makefiles and then the build process itself runs without error messages (which doesn´t seem to mean it runs at it should). Because then it runs into an error after make tht the file lib/libh3* doesn´t exist. (which is true after the build process)
So just for trying I´ve tried to swap lib/libh3* by bin/libh3* and after the shell script runs complete without any error. However no python package is created and if there is any further step I just don´t know how to continue from there.

techshot25:
If you really want h3 on your windows machine without any problems, you can install everything related to your python environment inside a linux subsystem which may seem like a lot of work, but it allows you to utilize every piece of software that was built for linux and works exclusively for it. I simply moved my entire work to the linux subsystem and it has never failed me.

MustaphaAmine:
Can you please tell me how to install h3 on Linux subsystem. because I've tried every possible way to install it on my windows machine but without success.

ajfriend:
We just released v3.6.1, which should hopefully fix these install issues. Could you please try installing with `pip install h3==3.6.1` and let us know how it works?

ajfriend:
Closing for now, please re-open if the issue is still there.

chisomeluwa:
It still does not work on Windows @ajfriend, I get the error 
 _" note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for h3
Failed to build h3
ERROR: Could not build wheels for h3, which is required to install pyproject.toml-based _projects"__

ajfriend:
[...]

PR #214, opened 2022-02-22
Run CI Tests on 3.10, remove Python 3.5 Windows test/wheel

kylebarron:
It seems from #213 that some CI runs are failing that shouldn't be related to that PR's changes, e.g. tests and wheel builds on Windows on Python 3.5. I'm using this PR to test this apart from any other changes.

### Change list

- Run tests on Python 3.10
- Don't run tests on Python 3.5 on Windows
- Don't build wheel for Python 3.5 on Windows

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
+        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9, "3.10"]
         exclude:
           - os: windows-latest
             python-version: 2.7
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,6 +18,8 @@ jobs:
         exclude:
           - os: windows-latest
             python-version: 2.7
+          - os: windows-latest
+            python-version: 3.5
 
     steps:
       - uses: actions/checkout@v2.3.3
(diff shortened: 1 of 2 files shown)


kylebarron:
Yeah this PR is failing because building h3 on 3.5 on Windows no longer works. https://github.com/uber/h3-py/runs/5295628853?check_suite_focus=true

ajfriend:
Hmm. I wonder what changed to break 3.5 on windows? I'm curious because maybe we're forgetting to pin a version somewhere.

kylebarron:
I don't know. I know nothing about Windows compiling setups. My first guess is that we're using `windows-latest` and maybe the OS was updated?

dfellis:
Python 3.5 got EOLed last year https://endoflife.date/python

Perhaps Github is starting to remove support for it in their runners?

ajfriend:
Fair enough.

Related, I would like to move to [`cibuildwheel >= 2.0`](https://cibuildwheel.readthedocs.io/en/stable/changelog/#v200), but we don't have to do that in this PR, unless you're really motivated @kylebarron. :)

Just calling it out because it removes support for 2.7 and 3.5, so we'd probably drop 3.5 altogether. But I'm thinking of having a separate action using `cibuildwheel < 2.0` to still build the 2.7 wheels if that doesn't cause us too much extra trouble.

But for now, this looks great!

ajfriend (approved):

kylebarron:
I think a separate PR would be better for moving to 2.x.

The main thing keeping us on cibuildwheel 1.x seems to be the Python 2.x support. As you know I'd love to deprecate support for Python 2 but I'm biased 😄 .

kylebarron:
It also looks like you can't build wheels for Python 3.10 using cibuildwheel 1.x. So we might get more issues like [this one](https://github.com/uber/h3-py/issues/18#issuecomment-1047725349) asking why they can't install on Python 3.10.

ajfriend:
I'd also like to drop 2.7, but I think there's still a nontrivial population that's stuck with it in the GIS community (but I may be out of date on that).

PR #215, opened 2022-02-23
Add auto formatter

kylebarron:
As I learned in #213 , h3 currently uses `flake8` for linting, requiring a max line length of 88 chars, but does not use an auto-formatting solution. This means that manual editing is required to format code in a way that passes `flake8`.

This PR adds two autoformatters: `black` and `isort`. [`black`](https://github.com/psf/black) is likely the most commonly used Python formatter at this time and is [used by many popular projects](https://github.com/psf/black#used-by). `black` autoformats code while [`isort`](https://github.com/PyCQA/isort) sorts imports deterministically. This PR uses [`pre-commit`](https://pre-commit.com/) as a method to easily run all linting together, pinned to specific tag versions. And it can run automatically on each git commit if you run `pre-commit install`.

Formatting is bound to be subjective, so I'm happy to hear criticism or changes desired, but I personally would be in favor of _any_ solution that avoids the need for manual editing of code to meet formatting requirements.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,31 @@
...
+    rev: 5.10.1
+    hooks:
+      - id: isort
+
+  - repo: https://github.com/psf/black
+    rev: 22.1.0
+    hooks:
+      - id: black
+        args: [
+          --skip-string-normalization
```
kylebarron, line 25: This argument is added for a smaller diff. The project historically uses single quotes for strings, while black will convert all strings to double quotes for consistency.

isaacbrodsky (approved):

ajfriend:
I'll try https://github.com/uber/h3-py/issues/337

PR #216, opened 2022-02-24
Add type hints to api functions class

kylebarron:
### Changes

- Make this `_API_FUNCTIONS` class inherit from `Generic` to support its use as a [generic class](https://mypy.readthedocs.io/en/stable/generics.html#defining-generic-classes). This solves the typing problem without needing to keep track of type hints defined separately from code and with no duplication (see closed PR #194).

#### Backwards compatible with Python 2.7

Uses type hints inside comments for backwards compatibility at runtime. If/when we deprecate support for Python 2.7 we can move the type hints into function parameters. All existing tests are passing.

#### Typing without separate stub files

Defines `_API_FUNCTIONS` as a generic class, which can be parametrized with different scalar types. This allows mypy to understand function typing. 

```py
from h3.api import basic_int
from typing import List, Set

out: Set[int] = basic_int.k_ring(1, 2)
# Passes

out2: Set[str] = basic_int.k_ring(1, 2)
# error: Incompatible types in assignment (expression has type "Set[int]", variable has type "Set[str]")

out3: List[int] = basic_int.k_ring(1, 2)
# error: Incompatible types in assignment (expression has type "Set[int]", variable has type "List[int]")

out4 = basic_int.k_ring('a', 2)
# error: Argument 1 to "k_ring" of "_API_FUNCTIONS" has incompatible type "str"; expected "int"
```

### Todo

- [ ] Check some of the input/returned types (see a couple `TODO` comments in the code)
    - [ ] Discussion about whether the numpy int api should return numpy int objects: https://github.com/uber/h3-py/pull/213#discussion_r837198430
- [ ] Add type tests. See https://github.com/uber/h3-py/pull/194/files#diff-c02f137b8e703d2c3920f300d2929be20f20a5fed41cb453cf4c29221e3d7253 for reference
- [x] Add type hints to `h3.unstable.vect`.

Ref https://github.com/uber/h3-py/pull/194#issuecomment-1043368400


.github/workflows/tests.yml
```diff
@@ -12,6 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
+      fail-fast: false
```
kylebarron, line 15: This should be reverted before merging

kylebarron:
I still need to add type tests, but this should be good for an initial review

src/h3/api/numpy_int/_binding.py
```diff
@@ -21,13 +21,23 @@
 
 import numpy as np
 
+try:
+    from numpy.typing import NDArray
+    ArrayType = NDArray[np.uint64]
+except ImportError:
+    ArrayType = np.ndarray
```
kylebarron, line 28: The numpy.typing module is new in v1.20: https://numpy.org/devdocs/reference/typing.html

kylebarron:
Converting to a draft just to avoid building wheels on CI for now

kylebarron:
I started the test setup. I'll try to flesh out the actual tests later this week.

kylebarron:
It's pretty tedious to write all the type tests, so I still haven't finished them, but you can see the last three commits: https://github.com/uber/h3-py/pull/216/files/58f5e4f9dcc188a77eb0b33e4cc03dec82777d0a..22d2b3d7ecd25a4283be13fd71fec5b73781fad3 for how the type tests are written.

PR #218, opened 2022-03-10
Alternate tooling support approach

ajfriend:
@kylebarron sorry for the delay on this, but this is roughly what I was thinking. Still a lot to clean up.

One of the reasons I like this approach is that it avoids the code generation step.

I'm not sure when/if we'd attack the type information, but if we did, how would this approach mesh with what you were thinking?


kylebarron:
To be honest I'm not sure I see the appeal of this approach. This basically takes away all the DRY? We need to manually ensure that code changes, docstring changes, and type hinting changes are manually copied between the APIs, which seems like a lot of work? The test with `inspect.getsource` asserts that the textual source of every API is the same, but then 

The point of #213 was to maintain the benefits of DRY for the API definitions while enabling static tooling. To not need to copy all of `_copy_of_api` four times, while having clean typing through generics.

In terms of typing, it might still be possible with this approach. `inspect.getsource` checks the text, not the AST, of the input, so we couldn't have any API-specific typing literals in the code. We could in theory have type aliases in the same file as `inputs_outputs`, but it's still more verbose.

The code generation in #213 is quite small; it just binds names to the scope statically.
```
cell_area = basic_int.cell_area
compact = basic_int.compact
...
```
If you really don't like that code generation, we could take a more manual approach like this PR where the binding is manual and there's only the code that asserts.

ajfriend:
Good to hear that if we were to do this, it wouldn't block the typing.

The tradeoff I'm considering is the added complexity of the code generation vs the need for manual syncing of the APIs. This is definitely the lower tech solution, but I don't think the manual syncing is actually going to be that much extra work in practice. Someone would just change the code in one of the API files and copy the whole file over to the others. You wouldn't need to make sure they're the same item by item, it's just a wholesale cut-and-paste.

You could even have a script to do it. Maybe there's a halfway point between this and #213 that does that copying automatically?

ajfriend:
Also, I certainly don't think of this as the ultimate, perfect solution to this problem. But it did seem like a way to get that IDE functionality that you were looking for and I think a lot of users would appreciate in a quick way that doesn't change the current flow too much.

And if we can unblock that functionality in the short term for the 3x branch, I'd prefer to spend more time looking forward to 4.0. Since 4.0 will be in active development, I think that might be an even better place to explore different approaches to this problem.

ajfriend:
@dfellis made the good suggestion that we could DRY up this approach using symbolic links, which are supported by git.

kylebarron:
> The tradeoff I'm considering is the added complexity of the code generation vs the need for manual syncing of the APIs. 

I don't really share the worries of complexity of code generation, because the output isn't hidden -- the output is checked in to Git along with a PR. I don't have strong preferences over manually syncing vs generating _binding names_.

However I do feel strongly that we should not be manually copying entire function definitions four times. I think that would make the codebase more unwieldy and increase developer frustration.

I updated #213 to remove any code generation, while manually copying only the 60 lines of code in each `_public_api.py`. This seems to check all the boxes?

- [x] No code generation
- [x] No dynamic population of `globals`
- [x] Full DRY with no function definition copied
- [x] Tooling support
- [x] Easier type hinting that only needs to be defined once, in `_api_template` instead of four times in each bindings file.

Is there something about the class-based approach you don't like?

PR #220, opened 2022-04-05
Set up hybrid cibuildwheel v1 + v2 wheel builds, add Python 3.10 wheels

kylebarron:
We want to adopt cibuildwheel v2 for support for Python 3.10 and potentially other platforms, but cibuildwheel v2 dropped support for Python 2.7. This PR sets up the CI to use both cibuildwheel versions.

In a future PR we could likely explore building apple silicon wheels using cibw v2.

Closes https://github.com/uber/h3-py/issues/18

### DIFF
--- a/.github/workflows/coverage-lint.yml
+++ b/.github/workflows/coverage-lint.yml
@@ -1,10 +1,10 @@
 name: coverage-lint
 
-on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
+# on:
+#   push:
+#     branches: [master]
+#   pull_request:
+#     branches: [master]
 
 jobs:
   tests:
--- a/.github/workflows/coverage-lint.yml
+++ b/.github/workflows/coverage-lint.yml
@@ -1,10 +1,10 @@
 name: coverage-lint
 
-# on:
-#   push:
-#     branches: [master]
-#   pull_request:
-#     branches: [master]
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
 
 jobs:
   tests:
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -9,14 +9,12 @@ on:
     types:
       - published
 
-
 jobs:
   make_sdist:
     name: SDist
     runs-on: ubuntu-latest
 
     steps:
-
       - uses: actions/checkout@v2
         with:
           submodules: recursive
@@ -30,7 +28,6 @@ jobs:
         with:
           path: ./dist
 
-
   make_cibw_v1_wheels:
     name: "cibuildwheel v1: ${{ matrix.name }}"
     runs-on: ${{ matrix.os }}
@@ -39,18 +36,17 @@ jobs:
       matrix:
         # Only build for platforms that cibw v2 doesn't support
         include:
-        - os: macos-latest
-          build: "cp27-macosx_x86_64"
-          name: macOS Intel
-        - os: windows-latest
-          build: "cp27-win_amd64"
-          name: Windows 64-bit
-        - os: ubuntu-latest
-          build: "cp27-manylinux_x86_64"
-          name: Linux Intel 64-bit
+          - os: macos-latest
+            build: "cp27-macosx_x86_64"
+            name: macOS Intel
+          - os: windows-latest
+            build: "cp27-win_amd64"
+            name: Windows 64-bit
+          - os: ubuntu-latest
+            build: "cp27-manylinux_x86_64"
+            name: Linux Intel 64-bit
 
     steps:
-
       - uses: actions/checkout@v2
         with:
           submodules: recursive
@@ -67,12 +63,6 @@ jobs:
           echo "CC=cl.exe" >> $GITHUB_ENV
           echo "CXX=cl.exe" >> $GITHUB_ENV
 
-      - name: Set up QEMU
-        if: runner.os == 'Linux'
-        uses: docker/setup-qemu-action@v1
-        with:
-          platforms: aarch64
-
       ## Build
 
       - uses: pypa/cibuildwheel@v1.12.0
@@ -81,7 +71,6 @@ jobs:
           CIBW_TEST_COMMAND: pytest {project}/tests
           CIBW_ARCHS_LINUX: auto aarch64
           CIBW_BUILD: ${{ matrix.build }}
-          CIBW_SKIP: ${{ matrix.skip }}
 
       - name: Check with Twine
         run: |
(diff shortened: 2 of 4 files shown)


.github/workflows/coverage-lint.yml
```diff
@@ -1,10 +1,10 @@
 name: coverage-lint
 
-on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
+# on:
```
kylebarron, line 3: Other CI commented out for now until wheels workflows are working

.github/workflows/wheels.yml
```diff
@@ -31,18 +31,79 @@ jobs:
...
+        shell: bash
+        run: |
+          echo "CC=cl.exe" >> $GITHUB_ENV
+          echo "CXX=cl.exe" >> $GITHUB_ENV
+
+      - name: Set up QEMU
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: aarch64
```
kylebarron, line 74: I'm guessing this is only relevant for aarch64 builds, and so can be removed for cibw v1 wheels.

.github/workflows/wheels.yml
```diff
@@ -30,38 +28,89 @@ jobs:
...
     strategy:
+      fail-fast: false
       matrix:
+        # Only build for platforms that cibw v2 doesn't support
         include:
-        - os: macos-latest
-          build: "cp*-macosx_x86_64"
-          name: macOS Intel
-        - os: windows-latest
-          build: "cp3*-win_amd64"
```
kylebarron, line 44: ohhhh it's because we weren't building windows 2.7 wheels before

kylebarron:
Here's the diff between the old and new wheel artifacts from CI. I assume that the manylinux names are equivalent. So in effect this currently:

- removes manylinux 3.5
- adds Python 3.10 on mac, windows, and linux.

If manylinux 3.5 is important, we could add that to cibw_v1. If Mac 3.5 is important, we could also add that to cibw_v1, but need to downgrade the mac runner. There's this warning in CI:

> CPython 3.5 is unsupported when building on macOS 11. To build CPython 3.5 wheels, build on an older OS, such as macOS 10.15. To silence this warning, deselect CPython 3.5 by adding "cp35-macosx_x86_64" to your CIBW_SKIP option.

```diff
4c4,6
< h3-3.7.3-cp35-cp35m-manylinux2010_x86_64.whl
---
> h3-3.7.3-cp310-cp310-macosx_10_9_x86_64.whl
> h3-3.7.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
> h3-3.7.3-cp310-cp310-win_amd64.whl
6,7c8,9
< h3-3.7.3-cp36-cp36m-manylinux2010_x86_64.whl
< h3-3.7.3-cp36-cp36m-manylinux2014_aarch64.whl
---
> h3-3.7.3-cp36-cp36m-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
> h3-3.7.3-cp36-cp36m-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
10,11c12,13
< h3-3.7.3-cp37-cp37m-manylinux2010_x86_64.whl
< h3-3.7.3-cp37-cp37m-manylinux2014_aarch64.whl
---
> h3-3.7.3-cp37-cp37m-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
> h3-3.7.3-cp37-cp37m-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
14,15c16,17
< h3-3.7.3-cp38-cp38-manylinux2010_x86_64.whl
< h3-3.7.3-cp38-cp38-manylinux2014_aarch64.whl
---
> h3-3.7.3-cp38-cp38-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
> h3-3.7.3-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
18,19c20,21
< h3-3.7.3-cp39-cp39-manylinux2010_x86_64.whl
< h3-3.7.3-cp39-cp39-manylinux2014_aarch64.whl
---
> h3-3.7.3-cp39-cp39-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
> h3-3.7.3-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

isaacbrodsky (approved):

dfellis (approved):

kylebarron:
I removed `fail-fast: false`. That's really nice for testing the CI workflow file, because with one bug you can still see if the rest of the workflows are running fine. But I _think_ it's not ideal for production because you can saturate the concurrent number of builds more easily with `fail-fast: false`.

ajfriend:
> If manylinux 3.5 is important, we could add that to cibw_v1. If Mac 3.5 is important, we could also add that to cibw_v1, but need to downgrade the mac runner. There's this warning in CI:

I'm cool with dropping support for `3.5`.

ajfriend:
> I removed `fail-fast: false`. That's really nice for testing the CI workflow file, because with one bug you can still see if the rest of the workflows are running fine. But I _think_ it's not ideal for production because you can saturate the concurrent number of builds more easily with `fail-fast: false`.

Sound good. I think it's probably just a remnant of working through a bug on one platform or another.

.github/workflows/wheels.yml
```diff
@@ -30,38 +28,72 @@ jobs:
...
+            name: Linux Aarch64 3.6
+          - os: ubuntu-latest
+            build: "cp37-manylinux_aarch64"
+            name: Linux Aarch64 3.7
+          - os: ubuntu-latest
+            build: "cp38-manylinux_aarch64"
+            name: Linux Aarch64 3.8
+          - os: ubuntu-latest
+            build: "cp39-manylinux_aarch64"
+            name: Linux Aarch64 3.9
```
ajfriend, line 94: Does this simplify if we give up `3.5` support?

ajfriend:
Just noticing while we're here: Either for this or a follow-up PR, let's try upgrading to `docker/setup-qemu-action@v1.2.0`

.github/workflows/wheels.yml
ajfriend: That is, could we now we get away with just something like:

```
  build_wheels:
    name: Build wheels on ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-20.04, windows-2019, macos-10.15]
```

as in https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions ?

kylebarron: I think it could potentially simplify. There's also a tradeoff here between simplicity and parallelizing the runners. If you break out more matrix options here, you're able to better parallelize the builds.

My idea for this PR was to update to v2 without making substantive changes, which can then be addressed in future PRs

kylebarron:
> Sound good. I think it's probably just a remnant of working through a bug on one platform or another.

😅 oh I wasn't clear; it isn't currently set on master. I was just commenting that I reverted those changes on this branch 😄

kylebarron:
> Just noticing while we're here: Either for this or a follow-up PR, let's try upgrading to `docker/setup-qemu-action@v1.2.0`

👍  I updated this pin to 1.2.0

ajfriend (approved):

.github/workflows/wheels.yml
kylebarron: > Does this simplify if we give up `3.5` support?

I'm not sure I follow this (or maybe you meant `2.7`?). cibuildwheel v2 doesn't support Python 3.5: https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip

So any 3.5 wheels would still need to be built with cibuildwheel v1.

Issue #205, opened 2021-11-09
Should the vectorized geo_to_h3 stay as unstable?


thomasaarholt:
#147 introduced vectorized support for geo_to_h3, which sped things up by about 50% over a python for loop. It's currently under `unstable`, which raises a warning. It seems that `geo_to_h3` gives accurate results, and that its API won't retroactively break in the future. @ajfriend do you agree?

Can I suggest that the functionality gets moved to `api` instead of `unstable`? To me, the warning from `unstable` implied that it may not give correct results, which does not seem to be the case.

ajfriend:
Maybe we should update the warning, as `unstable` was meant to imply that the API will probably change in the future. (Happy to accept a PR on this, btw!)

Also, we're planning on a 4.0 release when the core library also moves to 4.0, so many of the function names will change anyway and break backwards compatibility. For 4.0, I would like to settle on how exactly we'd like to implement the vectorized functions. For example, it'd be nice if we could have them work like `numpy` functions and dispatch correctly to both vector and scalar arguments. (Also happy to accept PRs to play around in the `unstable` namespace to try out different approaches for this.)

kylebarron:
> For 4.0, I would like to settle on how exactly we'd like to implement the vectorized functions.

I think 4.0 is coming up pretty soon 😄 . I'd love to get more of the H3 API to have a vectorized counterpart. My current thoughts on this are:

- Add more `.pxd` files to prevent duplicating Cython wrapper code for vectorized functions. The existing approach for the few vectorized functions calls into `h3lib` directly. https://github.com/uber/h3-py/blob/034a41e6237a850dec7f984963d87603c9342810/src/h3/_cy/unstable_vect.pyx#L59-L60 This isn't horrible for simple functions like `geo_to_h3`, but it's not ideal for other functions that require more Cython wrapper code around `h3lib`. Adding `.pxd` definition files (that define headers for our "exported" `cpdef` functions) for our `.pyx` implementation files will allow us to import those `.pyx` files from other Cython files. (This would also be relevant for https://github.com/uber/h3-py/issues/152).

    So instead of calling `h3lib` directly, we'd call our own internal wrapper functions:

	```py
	from .geo import geo_to_h3
	
	...
	with nogil:
	    for i in range(len(lat)):
	        out[i] = geo_to_h3(lat[i], lng[i])
	```

[...]

Issue #152, opened 2020-07-16
Example using `cimport h3` in external code for speed


ajfriend:

kylebarron:
Just a note I'm guessing there would need to be more `.pxd` files in order for the end-user `cimport` to work. In Jupyter:

```
> ipython
Python 3.8.3 (default, May 19 2020, 13:54:14)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.20.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %load_ext Cython
   ...:

In [2]: %%cython
   ...: cimport h3
   ...:
   ...:

Error compiling Cython file:
------------------------------------------------------------
...
cimport h3
       ^
------------------------------------------------------------

/Users/kyle/.ipython/cython/_cython_magic_f4fb333f094afe9fe31d61aa9102128e.pyx:1:8: 'h3.pxd' not found
```

Does that work for you?

kylebarron:
It seems like this should work by adding `.pxd` definition files. Anything `cimport`'ed needs to have a corresponding `.pxd`. One note is that any code that calls `cimport h3` looks like it'll need a path to `h3api.h` as well

```
> ipython
Python 3.8.3 (default, May 19 2020, 13:54:14)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.20.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %load_ext cython

In [2]: %%cython
   ...: from h3._cy cimport util
   ...:
   ...:
/Users/kyle/.ipython/cython/_cython_magic_2b223eeb87eac923ee742a354e855a5a.c:625:10: fatal error:
      'h3api.h' file not found
#include "h3api.h"
         ^~~~~~~~~
1 error generated.
```

This was run in the `h3-py` root folder, letting the import find the package by file path. I think that the `.pxd` files might not be distributed as package data, so it won't work with the pip-installed `h3-py`.

Maybe it would be possible to define files like `h3api.h` and `.pxd` as package data included in the build?

ajfriend:
Makes sense. My impression was that this should be possible theoretically, but I don't know if I've ever seen any smaller projects implement it and it get used elsewhere. Probably just `numpy`, but that's a hard code base to cheat off of :)

Let me know if you get any further with this! I'll try to revisit it at some point, but if you beat me to it with a PR, that'd be great too!

kylebarron:
[...]

PR #224, opened 2022-04-11
Add `.pxd` defs to expose Cython functions

kylebarron:
Opening this draft PR for discussion around exposing Cython functions. 

Relevant for:

- Vectorized API with maximal code reuse: see https://github.com/uber/h3-py/issues/205#issuecomment-1095403427
- Public Cython API: https://github.com/uber/h3-py/issues/152

### DIFF
--- a/src/h3/_cy/cells.pyx
+++ b/src/h3/_cy/cells.pyx
@@ -130,10 +130,10 @@ cpdef H3int[:] ring(H3int h, int k):
     return mv
 
 
-cpdef H3int parent(H3int h, res=None) except 0:
+cpdef H3int parent(H3int h, int res = -1) except 0:
     check_cell(h)
 
-    if res is None:
+    if res == -1:
         res = resolution(h) - 1
     if res > resolution(h):
         msg = 'Invalid parent resolution {} for cell {}.'
--- a/src/h3/_cy/cells.pyx
+++ b/src/h3/_cy/cells.pyx
@@ -237,7 +237,7 @@ cpdef int64_t num_hexagons(int resolution) except -1:
     return h3lib.numHexagons(resolution)
 
 
-cpdef double mean_hex_area(int resolution, str unit='km^2') except -1:
+cpdef double mean_hex_area(int resolution, unit='km^2') except -1:
     check_res(resolution)
 
     area = h3lib.hexAreaKm2(resolution)
@@ -256,7 +256,7 @@ cpdef double mean_hex_area(int resolution, str unit='km^2') except -1:
     return area
 
 
-cpdef double cell_area(H3int h, str unit='km^2') except -1:
+cpdef double cell_area(H3int h, unit='km^2') except -1:
     check_cell(h)
 
     if unit == 'rads^2':
--- a/src/h3/_cy/cells.pyx
+++ b/src/h3/_cy/cells.pyx
@@ -130,10 +130,10 @@ cpdef H3int[:] ring(H3int h, int k):
     return mv
 
 
-cpdef H3int parent(H3int h, int res=-1) except 0:
+cpdef H3int parent(H3int h, res=None) except 0:
     check_cell(h)
 
-    if res < 0:
+    if res is None:
         res = resolution(h) - 1
     if res > resolution(h):
         msg = 'Invalid parent resolution {} for cell {}.'
@@ -144,10 +144,10 @@ cpdef H3int parent(H3int h, int res=-1) except 0:
 
     return h3lib.h3ToParent(h, res)
 
-cpdef H3int[:] children(H3int h, int res=-1):
+cpdef H3int[:] children(H3int h, res=None):
     check_cell(h)
 
-    if res < 0:
+    if res is None:
         res = resolution(h) + 1
     if res < resolution(h):
         msg = 'Invalid child resolution {} for cell {}.'
@@ -164,10 +164,10 @@ cpdef H3int[:] children(H3int h, int res=-1):
 
     return mv
 
-cpdef H3int center_child(H3int h, int res=-1) except 0:
+cpdef H3int center_child(H3int h, res=None) except 0:
     check_cell(h)
 
-    if res < 0:
+    if res is None:
         res = resolution(h) + 1
     if res < resolution(h):
         msg = 'Invalid child resolution {} for cell {}.'
--- a/src/h3/_cy/cells.pyx
+++ b/src/h3/_cy/cells.pyx
@@ -130,10 +130,10 @@ cpdef H3int[:] ring(H3int h, int k):
     return mv
 
 
-cpdef H3int parent(H3int h, res=None) except 0:
+cpdef H3int parent(H3int h, int res=-1) except 0:
     check_cell(h)
 
-    if res is None:
+    if res < 0:
         res = resolution(h) - 1
     if res > resolution(h):
         msg = 'Invalid parent resolution {} for cell {}.'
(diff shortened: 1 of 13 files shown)


ajfriend:
Sweet! My thinking was that if we get this working, we can remove the `unstable` API from `h3-py`, since that kind of experimentation should be very easy to do in other repos. And the complexity of those other repos should be low, because `h3-py` will have already handled all the complexity of building/linking the core C lib.

kylebarron:
I agree that the 4.0 release is too soon to decide on a public vectorized API. But in the long term I'd be in favor of having a vectorized API in this canonical repo.

src/h3/_cy/cells.pyx
```diff
@@ -164,10 +164,10 @@ cpdef H3int[:] children(H3int h, res=None):
 
     return mv
 
-cpdef H3int center_child(H3int h, res=None) except 0:
+cpdef H3int center_child(H3int h, int res=-1) except 0:
```
kylebarron: A note for future discussion: Presumably typing every input argument would lead to faster cython code, so that the generated C code doesn't need to do a ton of type checking. So defining this arg as `int res=-1` would (I'm guessing) be faster than `res=None`.

But I'll revert this for now so that this PR has minimal changes

ajfriend:
> I agree that the 4.0 release is too soon to decide on a public vectorized API. But in the long term I'd be in favor of having a vectorized API in this canonical repo.

I agree that it'd be nice to have a standard vectorized API. I'm just saying we don't need to do the experimental stuff here once we've got this working.

ajfriend:
I'd also like to expose the C functions directly (from `h3lib.pxd`), if that's possible. I think I've tried in the past to call them from a separate Cython project (assuming they were good to go, since they had a `.pxd` file), but couldn't get it working...

kylebarron:
I think this PR currently works only for exposing these functions internally. In order to export a Cython API outside of the project, `.pxd` files need to exist after the `pip install`. Right now the only files that exist after an install (at least on my machine) are `.so` files:

```
> ls env/lib/python3.8/site-packages/h3/_cy/
__init__.py                        edges.cpython-38-darwin.so         unstable_vect.cpython-38-darwin.so
__pycache__                        geo.cpython-38-darwin.so           util.cpython-38-darwin.so
cells.cpython-38-darwin.so         to_multipoly.cpython-38-darwin.so
```

If you reference scipy, they have `.pxd` files like [such](https://github.com/scipy/scipy/blob/main/scipy/optimize.pxd), and then you can import those easily:

```py
In [1]: %load_ext Cython

In [3]: %%cython
   ...: cimport scipy.optimize
   ...:
   ...:

# worked
```

ajfriend:
> I think this PR currently works only for exposing these functions internally. In order to export a Cython API outside of the project, `.pxd` files need to exist after the `pip install`. Right now the only files that exist after an install (at least on my machine) are `.so` files:

Ah, ok. That makes sense. So maybe we can include these in the manifest? or do we need to do an extra step to copy them over?

kylebarron:
> So maybe we can include these in the manifest? or do we need to do an extra step to copy them over?

I'm not sure... that's what got me started on the Manifest rabbit hole, hence making #227 . But it looks like pxd files are included in the manifest... so I'm not sure exactly where the pxd files are removed. It might be something specific to the cmake/scikit-build setup.

Another thing I noticed, the same as [this comment](https://github.com/uber/h3-py/issues/152#issuecomment-810709398), is `h3api.h` unable to be found. Is `h3api.h` generated? It looks like the C lib only has a [`h3api.h.in`](https://github.com/uber/h3/tree/master/src/h3lib/include)

dfellis:
Yeah it's generated from the `.in` file.

kylebarron:
I added a couple commits to show how the vectorized functions could call our internal cython API instead of calling in to `h3lib` directly. But note this benchmark comparing against the master branch:

**Calling internal cython functions** (commit [`798da86` (#224)](https://github.com/uber/h3-py/pull/224/commits/798da8653c587d8eb411ba7d0120d8fc430e57e5) )
```
geo_to_h3 bench
76.5 ms ± 941 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
h3_to_parent bench
6.52 ms ± 111 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
h3_get_resolution bench
2.39 ms ± 11.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

**On master branch (commit https://github.com/uber/h3-py/commit/d1a5a3b9b2add6f970a840f154c51947d312cdd7)**
```
geo_to_h3 bench
75.8 ms ± 611 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
h3_to_parent bench
371 µs ± 52.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
h3_get_resolution bench
188 µs ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

`geo_to_h3` is the same speed in each, but `h3_to_parent` and `h3_get_resolution` is roughly an order of magnitude on this branch at [`798da86` (#224)](https://github.com/uber/h3-py/pull/224/commits/798da8653c587d8eb411ba7d0120d8fc430e57e5).

I _believe_ the reasons for this is
1.  the Cython `geo_to_h3` signature is fully typed, while the `parent` signature is not. Additionally, the `parent` function accepts Python, not C types.

	https://github.com/uber/h3-py/blob/d1a5a3b9b2add6f970a840f154c51947d312cdd7/src/h3/_cy/geo.pyx#L17
	https://github.com/uber/h3-py/blob/d1a5a3b9b2add6f970a840f154c51947d312cdd7/src/h3/_cy/cells.pyx#L133

2. The `parent` and `resolution` functions do extra checks with `check_cell` and `check_res`.

I think it will be valuable in terms of code reuse for the vectorized API to call the internal Cython functions. So to keep performance high, I'd advocate for the Cython API to be fully typed. I.e.:

- Instead of `res = None`, use `res = -1`
- Create exception messages in Python instead of Cython?
- Analyze the Cython report to see where it's calling into Python

If we also remove any Python types from Cython, our Cython API can be marked as `nogil`, which would allow our vectorized API to release the GIL.

<details>
<summary>Bench script</summary>

Note this needs to be saved as an `.ipy` file and run with `ipython bench.ipy`

```py
import h3.unstable.vect as vect
import numpy as np

N = 100_000
K = 10

lats, longs = np.random.uniform(0, 90, N), np.random.uniform(0, 90, N)
h3_index = vect.geo_to_h3(lats, longs, 9)

print('geo_to_h3 bench')
%timeit vect.geo_to_h3(lats, longs, 9)

print('h3_to_parent bench')
%timeit vect.h3_to_parent(h3_index, 8)
# 411 µs ± 16.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

print('h3_get_resolution bench')
%timeit vect.h3_get_resolution(h3_index)
```
</details>

kylebarron:
Following up on the importance of typing cython function signatures, here's a before/after of this commit [`2bab57f` (#224)](https://github.com/uber/h3-py/pull/224/commits/2bab57f96831c200f532a9b527de6d75dfaf80c7), which just defined `res` as `int`.

When you look at the **before** Cython annotation, you can see that Cython needs to talk to Python for the `>` comparison (darker yellow is more interaction with the Python interpreter)

image

When you look at **after**, you can see that checking `res > resolution(h)` can be done fully statically. Additionally, `check_res` needs less interaction with Python, because Python is no longer needed to check that the input to `check_res` is in fact an `int`.

image

Now using the same bench script as the last commit, `h3_to_parent` takes 28% less time!
```
h3_to_parent bench
4.69 ms ± 17.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

The point here is that if we want a vectorized API to use our Cython API, and if we want that to be performant, we may need to make changes to our Cython API, and move some things out of Cython and up into Python.

kylebarron:
Closing as superseded by #233 and #234

PR #225, opened 2022-04-11
Musllinux wheels

kylebarron:
Seeing if the `cp*-musllinux_x86_64` wheels build out of the box. These wheels would be used in [Alpine linux](https://www.alpinelinux.org/)

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -79,7 +79,10 @@ jobs:
             name: Windows 64-bit
           - os: ubuntu-latest
             build: "cp*-manylinux_x86_64"
-            name: Linux Intel 64-bit
+            name: Linux Intel glibc 64-bit
+          - os: ubuntu-latest
+            build: "cp*-musllinux_x86_64"
+            name: Linux Intel musl 64-bit
           - os: ubuntu-latest
             build: "cp36-manylinux_aarch64"
             name: Linux Aarch64 3.6


kylebarron:
Looks like the musllinux wheels finished in 38 minutes

dfellis:
> Looks like the musllinux wheels finished in 38 minutes

38min vs 7min is crazy. I assume that's because of QEMU in there. Why not just Docker?

kylebarron:
> 38min vs 7min is crazy. I assume that's because of QEMU in there

🤷‍♂️  I'm not really familiar with the exact low-level differences between archs and glibc/musl... Though from the comment in the CI def, QEMU is only used for aarch64? So I assume these `musllinux` wheels don't use QEMU because these are x86_64 musllinux wheels. cibuildwheel also supports musllinux aarch64 wheels 😄 `cp*-musllinux_aarch64` but I don't know if that would be useful to build 😂

dfellis:
> > 38min vs 7min is crazy. I assume that's because of QEMU in there
> 
> man_shrugging I'm not really familiar with the exact low-level differences between archs and glibc/musl... Though from the comment in the CI def, QEMU is only used for aarch64? So I assume these `musllinux` wheels don't use QEMU because these are x86_64 musllinux wheels. cibuildwheel also supports musllinux aarch64 wheels smile `cp*-musllinux_aarch64` but I don't know if that would be useful to build.

Got it. It makes sense that it would be used for arm64 builds, but I'm still confused about the huge difference in build times between the two. If the "normal" ubuntu path is also using QEMU for arm64 builds there, they should have similar overheads?

dfellis:
I think I found the real reason why it takes so long: For the Alpine Linux path [it needs to build numpy from scratch](https://github.com/uber/h3-py/runs/5978853576?check_suite_focus=true#step:6:286) while I'm sure it's just downloading a pre-build wheel in the Ubuntu path.

kylebarron:
I see. Even though we don't compile against numpy headers, we require numpy for the tests
https://github.com/uber/h3-py/blob/034a41e6237a850dec7f984963d87603c9342810/.github/workflows/wheels.yml#L126

Numpy might distribute musllinux wheels soon: https://github.com/numpy/numpy/issues/20089. Interestingly, this is not a factor in the slowness of `manylinux_aarch64` wheels, because Numpy ships `manylinux_aarch64` wheels: https://pypi.org/project/numpy/#files

ajfriend:
@kylebarron @dfellis you both happy to land like this?

kylebarron:
lgtm.

dfellis (approved):

ajfriend (approved):

PR #227, opened 2022-04-12
Add `MANIFEST.in` file for smaller sdist

kylebarron:
By adding a `MANIFEST.in` file, we can exclude files not necessary to build. This reduces the sdist size from 17MB to 198KB.

You can test that this sdist covers the necessary files by installing from this sdist in a new virtualenv (note that I don't think this is actually tested right now in CI, that you can actually install from the sdist):

```
> pipx run build --sdist
> virtualenv test_env
> source ./test_env/bin/activate
> pip install dist/h3-3.7.3.tar.gz
Processing ./dist/h3-3.7.3.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: h3
  Building wheel for h3 (pyproject.toml) ... done
  Created wheel for h3: filename=h3-3.7.3-cp38-cp38-macosx_11_0_x86_64.whl size=690776 sha256=411c0831a2e966d139437d08712d97774c15b94e7c9960ebe45da4fa5455ab80
  Stored in directory: /Users/kbarron/Library/Caches/pip/wheels/de/7a/ec/181fcd0e08c7df8d1d6c52575ed2b2d61e42fbe7bb7cf1c2b0
Successfully built h3
Installing collected packages: h3
Successfully installed h3-3.7.3
```

Here's the full list of files copied into the sdist:

<details>

```
copying CHANGELOG.md -> h3-3.7.3
copying CMakeLists.txt -> h3-3.7.3
copying LICENSE -> h3-3.7.3
copying dev_notes.md -> h3-3.7.3
copying makefile -> h3-3.7.3
copying pyproject.toml -> h3-3.7.3
copying readme.md -> h3-3.7.3
copying requirements.in -> h3-3.7.3
copying setup.py -> h3-3.7.3
copying src/h3/CMakeLists.txt -> h3-3.7.3/src/h3
copying src/h3/__init__.py -> h3-3.7.3/src/h3
copying src/h3/_cy.pyx -> h3-3.7.3/src/h3
copying src/h3/_version.py -> h3-3.7.3/src/h3
copying src/h3.egg-info/PKG-INFO -> h3-3.7.3/src/h3.egg-info
copying src/h3.egg-info/SOURCES.txt -> h3-3.7.3/src/h3.egg-info
copying src/h3.egg-info/dependency_links.txt -> h3-3.7.3/src/h3.egg-info
[...]

PR #228, opened 2022-04-12
CI test that sdist can be installed

kylebarron:


kylebarron:
Interesting that the sdist test passes here.

ajfriend:
Can we also run the tests after the sdist install? In this and #227

kylebarron:
Seems easiest to just install from sdist in the tests

kylebarron:
We might as well merge this first and then update #227 from the master branch

ajfriend:
I like the idea of testing both install methods. Could you leave the `tests.yml` as it was? And then add the tests after building the sdist in `wheels.yml`?

I like that logical separation a bit better. That way, we're treating the sdist as just another "distribution" along with the wheels, and by adding the tests, we're testing them the same (since cibuildwheel runs the tests after building the wheel).

ajfriend:
Closed by #229

PR #229, opened 2022-04-13
Update wheels.yml

ajfriend:
Since I'm the one nit-picking #228 :)

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -35,8 +35,9 @@ jobs:
       - name: Install from sdist
         run: |
           pip install --upgrade pip setuptools wheel
-          pip install dist/h3-*.tar.gz[all]
-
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels.yml
```diff
@@ -31,6 +31,14 @@ jobs:
       - name: Make sdist
         run: |
           pipx run build --sdist
+          
+      - name: Install from sdist
+        run: |
+          pip install --upgrade pip setuptools wheel
+          pip install dist/h3-*.tar.gz[all]
```
kylebarron, line 38: I couldn't get it to work with both * and [all], hence why I first used the glob to rename the file

kylebarron (approved):

.github/workflows/wheels.yml
ajfriend: ah, ok!

PR #234, opened 2022-04-25
Expose cython API for external use

kylebarron:
### Change list

- Update `CMakeLists.txt` to include `.pyx` and `.pxd` files in distribution after skbuild is done processing
- Include `h3api.h` in final bundle. This can usually be found automatically when compiling, but if we have issues with this in the future, we can emulate Numpy and its [`np.get_include` function](https://numpy.org/devdocs/reference/generated/numpy.get_include.html). 
- Add test that uses the external Cython API.

Closes https://github.com/uber/h3-py/issues/152

### DIFF
--- a/.github/workflows/coverage-lint.yml
+++ b/.github/workflows/coverage-lint.yml
@@ -34,7 +34,10 @@ jobs:
         run: pylint --disable=all --enable=import-error tests/
 
       - name: Coverage
-        run: pytest --cov=h3 --full-trace --cov-report=xml
+        run: |
+          pip install Cython
+          cythonize -i tests/cython_example.pyx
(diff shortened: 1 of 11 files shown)


kylebarron:
Ah the test script in cibuildwheel needs to be updated to either ignore the Cython test or compile the cython file used in that test.

kylebarron:
I updated to ignore the test when the cython example wasn't compiled

ajfriend (approved):

ajfriend:
Sweeeeeeet!

ajfriend:
Really simple notebook example for future reference: https://gist.github.com/ajfriend/167afffde74bc08b0e809e10b92ecde3

PR #238, opened 2022-05-15
WIP: Starting work on v4, The Sequel

ajfriend:
I'm working on the transition from v3 to v4. With @kylebarron's changes in #234, I was planning on using a duplicate library `h3fake2` (since Python can't have two libraries sharing the same name) to temporarily replace the `h3lib` C functions in this repo. The idea would be that all Python tests would pass, and we can replace the code in `h3lib` with the new v4 C library. We can then transition each of the `h3-py` functions over to the v4 C library one at a time.

So this seems to be breaking when I build it locally, even though the code in this example works (but with the `h3fake2` library): https://gist.github.com/ajfriend/167afffde74bc08b0e809e10b92ecde3

@kylebarron @isaacbrodsky, any ideas what I'm doing wrong here?


ajfriend:
Currently breaking with the message:

```
  Error compiling Cython file:
  ------------------------------------------------------------
  ...
  from .h3lib cimport H3int, H3str, GeoCoord
  
  cimport h3fake2._cy.h3lib  # not working?
         ^
  ------------------------------------------------------------
  
  /tmp/pip-req-build-t8l6mppq/src/h3/_cy/util.pxd:3:8: 'h3fake2/_cy/h3lib.pxd' not found
```

I'm not sure what we need to do to make the `.pxd` file available?

kylebarron:
Seems to work locally for me. You should check if the pxd files exist in your local `site-packages`, and if not, you probably need to set the include path to include `h3fake2`

```
> pip install git+https://github.com/ajfriend/h3fake2.git
> ls env/lib/python3.8/site-packages/h3fake2/_cy
__init__.py                        edges.pxd                          h3lib.pxd
__pycache__                        edges.pyx                          to_multipoly.cpython-38-darwin.so
cells.cpython-38-darwin.so         geo.cpython-38-darwin.so           unstable_vect.cpython-38-darwin.so
cells.pxd                          geo.pxd                            util.cpython-38-darwin.so
cells.pyx                          geo.pyx                            util.pxd
edges.cpython-38-darwin.so         h3api.h                            util.pyx
```

```
> ipython
Python 3.8.10 (default, Jul 28 2021, 14:21:59)
Type 'copyright', 'credits' or 'license' for more information
IPython 8.3.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %load_ext Cython

In [2]: %%cython
   ...: cimport h3fake2._cy.h3lib
   ...:
   ...:

In [3]:
```

ajfriend:
Yeah, it does work for me locally when I bring it up in IPython or in a Jupyter notebook like the one above. 

It only fails to work when I'm trying to pip install h3.

@kylebarron, could you check out this branch and see if you get the same error when running `make init`?

kylebarron:
Yeah I get the same error. I assume you need to set the include path to h3fake2

ajfriend:
Oh, do you mean in the `src/h3/_cy/CMakeLists.txt`? What would that look like? I'm constantly mystified by cmake :/

kylebarron:
I don't know. I've never used cmake. With a basic Cython build setup, I have to pass in the path for numpy headers: https://github.com/kylebarron/pymartini/blob/4774549ffa2051c4ddaa254aca191bedcc4d30c5/setup.py#L55-L56

I assume it's the same for this in order to find the external `h3fake2`.

kylebarron:
I don't fully understand how `h3fake2` is supposed to make the upgrade process easier. Maybe there's an easier way to incrementally upgrade without making a new package?

ajfriend:
> I don't fully understand how `h3fake2` is supposed to make the upgrade process easier. Maybe there's an easier way to incrementally upgrade without making a new package?

I'm totally open to other ideas. This was just a way to allow us to migrate a single function at a time while maintaining a working package throughout the process.

kylebarron:
What about adding h3lib v4 as another submodule in this repo, instead of linking to a new repo?

ajfriend:
> What about adding h3lib v4 as another submodule in this repo, instead of linking to a new repo?

I tried that in a previous attempt, but couldn't figure out the cmake-foo to get that to work. If someone could get that working, that would be a great solution.

kylebarron:
> Edit: Actually, I think I had trouble getting two submodules from the same repo working.

You can just run `git submodule add [h3url]` in another directory. But you still need to deal with cmake because then you have duplicate build artifacts with the same name

kylebarron:
I think maybe it's easiest to start a new `v4` dev branch where the submodule is updated to h3 v4, and the bindings are commented out. Then one PR at a time, a function is uncommented while being updated to v4

ajfriend:
> I think maybe it's easiest to start a new `v4` dev branch where the submodule is updated to h3 v4, and the bindings are commented out. Then one PR at a time, a function is uncommented while being updated to v4

Yeah, it'll be something like that. But ideally, I'd like to not have to comment out all of the tests to start off with.

ajfriend:
So only the sdist job fails due to this fairly new `CMakeTests.cmake` file. We get an error like

```
  CMake Error at src/h3lib/CMakeLists.txt:458 (include):
    include could not find requested file:
  
      CMakeTests.cmake
```

I'm guessing I need to modify the `h3-py` CMake files, but it isn't immediately clear to me how.

isaacbrodsky (approved):
approving for merging into a dev branch

Issue #236, opened 2022-04-25
Discuss typing of Cython API


kylebarron:
Making a new issue to come back to what the Cython API should look like for best performance.

I added a couple commits to show how the vectorized functions could call our internal cython API instead of calling in to `h3lib` directly. But note this benchmark comparing against the master branch:

**Calling internal cython functions** (commit [`798da86` (#224)](https://github.com/uber/h3-py/pull/224/commits/798da8653c587d8eb411ba7d0120d8fc430e57e5) )
```
geo_to_h3 bench
76.5 ms ± 941 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
h3_to_parent bench
6.52 ms ± 111 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
h3_get_resolution bench
2.39 ms ± 11.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

**On master branch (commit https://github.com/uber/h3-py/commit/d1a5a3b9b2add6f970a840f154c51947d312cdd7)**
```
geo_to_h3 bench
75.8 ms ± 611 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
h3_to_parent bench
371 µs ± 52.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
h3_get_resolution bench
188 µs ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

`geo_to_h3` is the same speed in each, but `h3_to_parent` and `h3_get_resolution` is roughly an order of magnitude on this branch at [`798da86` (#224)](https://github.com/uber/h3-py/pull/224/commits/798da8653c587d8eb411ba7d0120d8fc430e57e5).

I _believe_ the reasons for this is
1.  the Cython `geo_to_h3` signature is fully typed, while the `parent` signature is not. Additionally, the `parent` function accepts Python, not C types.

	https://github.com/uber/h3-py/blob/d1a5a3b9b2add6f970a840f154c51947d312cdd7/src/h3/_cy/geo.pyx#L17
	https://github.com/uber/h3-py/blob/d1a5a3b9b2add6f970a840f154c51947d312cdd7/src/h3/_cy/cells.pyx#L133

2. The `parent` and `resolution` functions do extra checks with `check_cell` and `check_res`.

[...]

PR #240, opened 2022-05-21
v4: local ij functions to v4 cython

isaacbrodsky:


kylebarron (approved):

src/h3/api/_api_template.py
```diff
@@ -855,7 +855,7 @@ def experimental_h3_to_local_ij(self, origin, h):
         origin = self._in_scalar(origin)
         h = self._in_scalar(h)
 
-        i, j = _cy.experimental_h3_to_local_ij(origin, h)
+        i, j = _cy.cell_to_local_ij(origin, h)
```
ajfriend, line 858: Same here. I'd like to make no changes to this file at this stage.

We will change all these function names (and the broader API as well), just not at this stage of the migration.

ajfriend (changes requested):

ajfriend:
Also note that we're probably going to rework the Cython API (e.g., #236), so these functions will get significant changes later on anyway. So I'd rather defer the discussion of changes around these functions to the next stage.

ajfriend:
@isaacbrodsky, so I think I got you to entertain my admittedly idiosyncratic request to remove the Cython function name changes in

- `src/h3/_cy/cells.pxd`
- `src/h3/_cy/cells.pyx`

and the Python function name changes in

- `src/h3/_cy/__init__.py`
- `src/h3/api/_api_template.py`

ajfriend (approved):
lgtm! thanks!

PR #241, opened 2022-05-21
ajfriend/more v4 funcs

ajfriend:


src/h3/_cy/__init__.py
```diff
@@ -42,12 +47,12 @@
...
-    cell_area,
-    line,
+    # mean_hex_area,
+    # cell_area,
+    # line,
     # is_res_class_iii,
-    get_pentagon_indexes,
-    get_res0_indexes,
+    # get_pentagon_indexes,
+    # get_res0_indexes,
```
isaacbrodsky, line 55: I would just delete these lines. They are not planned to be uncommented.

src/h3/_cy/cells.pyx
```diff
@@ -281,87 +281,104 @@ cpdef int64_t num_hexagons(int resolution) except -1:
...
 
-#     area = h3lib.hexAreaKm2(resolution)
+    check_res(resolution)
 
-#     # todo: multiple units
-#     convert = {
-#         'km^2': 1.0,
-#         'm^2': 1000*1000.0
-#     }
+    err = h3lib.getHexagonAreaAvgKm2(resolution, &area)
```
isaacbrodsky, line 291: `except -1` is unused because this doesn't propagate `err`

src/h3/_cy/cells.pyx
```diff
@@ -281,87 +281,104 @@ cpdef int64_t num_hexagons(int resolution) except -1:
...
-#     check_cell(end)
+    return area
 
-#     n = h3lib.h3LineSize(start, end)
 
-#     if n < 0:
-#         s = "Couldn't find line between cells {} and {}"
-#         s = s.format(hex(start), hex(end))
-#         raise H3ValueError(s)
+cpdef H3int[:] line(H3int start, H3int end):
```
isaacbrodsky, line 326: I believe this needs `except *` to check for an exception?
I would suggest renaming this to `grid_path_cells` as we update the underlying C call.

src/h3/_cy/cells.pyx
kylebarron: `except -1` would still be used if something else failed, like `check_res`, right?

isaacbrodsky: In theory yes, but I don't think this function will return `-1`.

src/h3/_cy/__init__.py
ajfriend: My plan (which I probably haven't communicated well) was that once we finish the migration to remove `h2fake2` that this file will be unchanged compared to the v3 version on `master`. So eventually, yes, they will be uncommented.

I agree that all these function names should be changed to something better, I just want to avoid that to minimize the surface area of what we're changing at this step.

src/h3/_cy/cells.pyx
ajfriend: This is a great discussion to have, but my plan is to defer these till later when we can examine our strategy around exceptions as a whole, instead of a function-by-function basis.

That's why I want to minimize the changes we're making at this stage. I just want us to focus on removing `h3fake2` right now. That should be a fairly straightforward and mechanical process.

Error handling/naming/API will probably require more of an open discussion, so I'm hoping to handle those as their own separate issues/PRs.

src/h3/_cy/cells.pyx
ajfriend: See comment above. Totally agree we should have this discussion, but let's defer until we remove `h3fake2`.

src/h3/_cy/cells.pyx
```diff
@@ -281,87 +281,104 @@ cpdef int64_t num_hexagons(int resolution) except -1:
...
 
+    return mv
 
-# cpdef H3int[:] get_res0_indexes():
-#     n = h3lib.res0IndexCount()
 
-#     ptr = create_ptr(n)
-#     h3lib.getRes0Indexes(ptr)
-#     mv = create_mv(ptr, n)
+cpdef H3int[:] get_res0_indexes():
```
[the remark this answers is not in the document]
ajfriend: Agree. In next stage.

src/h3/_cy/__init__.py
ajfriend: And to be clear, this file will eventually change with better names, just not at this stage.

ajfriend:
@isaacbrodsky At a high level, for this stage, I want us to focus on removing `h3fake2` and avoid any changes that might spark reasonable discussion. I'd like this stage to be pretty much just a mechanical process of moving over to the v4 C functions.

I'd like to have those reasonable discussions once we're past this stage.

isaacbrodsky (approved):
Discussed offline, we’ll land this as an interim step

src/h3/_cy/cells.pyx
ajfriend: To be discussed in #243

src/h3/_cy/cells.pyx
ajfriend: To be discussed in #243

PR #248, opened 2022-05-25
get_faces

ajfriend:


ajfriend:
Note to consider the "sizeof" naming conventions in the C library.

```
    # names within pair don't match
    # count vs size in function names?
    # getX() function
    # should have a getX_size() function?
    # what about max_ prefix? should we just coordinate on *_size?
    # if a C lib function has a companion "size" function, should we just always append with `_size`? if it is "max" size or "exact" size? who cares? its just memory management
    # 'Max'/"size"/"count"
    # prefix makes the function's function more obvious?
    # conjunction junction!
    # postfix makes it lexicographically sort with the associated function? (i like this one)
    #
    # is 'getNumCells' a special case? or should it also conform?
```

nrabinowitz (approved):

isaacbrodsky (approved):

PR #252, opened 2022-05-26
Uncomment cython example, and exclude sdist from wheels test temporarilty

ajfriend:


.github/workflows/wheels.yml
```diff
@@ -168,7 +168,8 @@ jobs:
           path: wheelhouse/*.whl
 
   upload_all:
-    needs: [make_sdist, make_cibw_v1_wheels, make_cibw_v2_wheels]
+    # needs: [make_sdist, make_cibw_v1_wheels, make_cibw_v2_wheels]
+    needs: [make_cibw_v1_wheels, make_cibw_v2_wheels]
```
isaacbrodsky, line 172: Can this be fixed in #239 rather than commenting out here?

ajfriend: Yeah, we can get rid of this once we fix #239. But without this, the wheels github action currently fails because the `make_sdist` action doesn't exist (because it is commented out above).

PR #258, opened 2022-06-04
Build wheels from sdist

ajfriend:


.github/workflows/wheels.yml
```diff
@@ -19,83 +19,32 @@ on:
...
-      - name: Check with Twine
-        run: |
-          pipx run twine check wheelhouse/*
-
-      - name: Upload artifacts to GitHub
-        uses: actions/upload-artifact@v3
-        with:
-          path: wheelhouse/*.whl
-
   make_cibw_v2_wheels:
```
kylebarron, line 45: Nit: rename to `make_wheels` now that we don't use both v1 and v2?

.github/workflows/wheels.yml
```diff
@@ -19,83 +19,32 @@ on:
...
-        run: |
-          pip install --upgrade pip setuptools wheel
-          pip install pytest
-          cp dist/h3-*.tar.gz h3.tar.gz
-          pip install h3.tar.gz[all]
-
-      - name: Test sdist
-        run: pytest --cov=h3 --full-trace
+          pip install --upgrade pip setuptools
+          pip install --upgrade build
```
kylebarron, line 37: Install on one line so that pip can resolve dependencies jointly?

.github/workflows/wheels.yml
```diff
@@ -152,37 +101,48 @@ jobs:
...
+        uses: actions/setup-python@v3
+
       - uses: actions/download-artifact@v3
         with:
           name: artifact
           path: dist
 
+      - name: Check with Twine
+        run: |
+          pipx run twine check dist/*
```
kylebarron, line 144: Is `pipx` installed with `actions/setup-python`? Do you need to `pip install twine`, or does `pipx run` do that automatically if not installed?

kylebarron (approved):

.github/workflows/wheels.yml
ajfriend: My habit has been to upgrade pip and setuptools first, to ensure that we have the latest version of pip that actually includes the (fairly new) dependency resolver. Not sure if that's entirely necessary here, so I just have it this way to be safe. Also, to ensure that we're installing `build` using the latest version of `pip`.

And I don't think we'd have any dependency issues between these three packages anyway. At least, if there were, that'd be something we'd want to pin explicitly. Would you agree?

kylebarron: Interesting. That's never crossed my mind before. I've just defaulted to installing on one line, but makes sense that might use the resolver of the original pip version, not the new one. Seems fine to leave as is

ajfriend:
Possible interim backup plan: just use the command line version of the `cibuildwheel` tool, instead of the github action?

.github/workflows/wheels.yml
ajfriend: These are good questions! I haven't entirely figured out what exactly `actions/setup-python` does and doesn't do. If you have any idea, that'd be a big help. Otherwise, I'll try to figure this out eventually.

kylebarron:
What would using the CLI solve instead of the action? It would be surprising if there's something you can do with the CLI that you can't with the action

ajfriend:
Closing due to inactivity.

PR #259, opened 2022-06-04
Remove more sdist files

ajfriend:
I don't think there's any need for the following files to be included in the Python SDist:

```
src/h3lib/scripts
src/h3lib/examples
src/h3lib/src/apps
```

@isaacbrodsky, could you confirm that's true?

This will further reduce the SDist size (beyond the huge gains of #227), and may also decrease the build time (?).

However the build still fails, but I'm hoping that some small changes to the cmake files can avoid the need for these files?


ajfriend:
I tried setting

```
set(BUILD_FUZZERS OFF CACHE BOOL "" FORCE)
set(ENABLE_LIBFUZZER OFF CACHE BOOL "" FORCE)
set(ENABLE_COVERAGE OFF CACHE BOOL "" FORCE)
```

but that didn't seem to help.

kylebarron:
I think when I was stripping sdist size in https://github.com/uber/h3-py/pull/227, it didn't build out of the box if I removed those files. I figured the further gains were marginal so I didn't look more closely at how to make the build work after removing those.

isaacbrodsky:
Was the error the same when BUILD_FUZZERS was OFF? Looking at CMakeLists.txt my suspicion is that the targets for executables are disabled, but other references to the source files remain, for example for autoformatting.

ajfriend:
> Was the error the same when BUILD_FUZZERS was OFF? Looking at CMakeLists.txt my suspicion is that the targets for executables are disabled, but other references to the source files remain, for example for autoformatting.

Getting different error output when I set BUILD_FUZZERS to OFF in the diff above.

ajfriend:
OK, cool. Let's just turn off autoformatting. I don't think there's a need for that at this stage. How would we configure that?

ajfriend:
> OK, cool. Let's just turn off autoformatting. I don't think there's a need for that at this stage. How would we configure that?

nvm. Just discovered the `cmake -LA` command. :)

isaacbrodsky:
There's a reference at https://h3geo.org/docs/next/core-library/compilation-options

CMakeLists.txt
```diff
@@ -7,18 +7,22 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
+turn_off(BUILD_BENCHMARKS)
+turn_off(BUILD_FILTERS)
+turn_off(BUILD_FUZZERS)
+turn_off(BUILD_GENERATORS)
+turn_off(BUILD_TESTING)
+turn_off(ENABLE_COVERAGE)
+turn_off(ENABLE_DOCS)
+turn_off(ENABLE_FORMAT)
+turn_off(ENABLE_LIBFUZZER)
+turn_off(ENABLE_LINTING)
```
kylebarron, line 25: Just curious, can you compare compilation times before and after this change? Does this make it a lot faster?

kylebarron (approved):

CMakeLists.txt
```diff
@@ -7,11 +7,22 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
-set(BUILD_FILTERS OFF CACHE BOOL "" FORCE)
-set(BUILD_GENERATORS OFF CACHE BOOL "" FORCE)
-set(BUILD_TESTING OFF CACHE BOOL "" FORCE)
-set(ENABLE_DOCS OFF CACHE BOOL "" FORCE)
+# See all options with `cmake -LA` in an `h3/build` directory,
+# or at https://h3geo.org/docs/next/core-library/compilation-options/
+macro(turn_off option_name)
+    set(${option_name} OFF CACHE BOOL "" FORCE)
+endmacro()
+turn_off(BUILD_ALLOC_TESTS)
```
isaacbrodsky, line 15: BUILD_ALLOC_TESTS should not have any effect when BUILD_TESTING is off. Harmless to disable it though.

isaacbrodsky (approved):

CMakeLists.txt
ajfriend: I had a suspicion that that was true, but yeah, figured it couldn't hurt to just disable everything to make sure.

CMakeLists.txt
ajfriend: I spot checked some other compilation times. There's *maybe* a gain. I'll try some runs here for a clearer comparison.

ajfriend: Seems to be faster, based on the last two runs I looked at, but hard to tell how much is random variation.

**Old config**
Screen Shot 2022-06-06 at 3 51 11 PM

**New config**
Screen Shot 2022-06-06 at 3 51 40 PM

PR #260, opened 2022-06-05
New error system

ajfriend:
# notes

- This PR helped me spot and (I think) fix several memory allocation bugs. These bugs all followed the same pattern, and will be (or would have been) fixed by #261

### DIFF
--- a/src/h3/_cy/cells.pyx
+++ b/src/h3/_cy/cells.pyx
@@ -78,10 +78,10 @@ cpdef H3int[:] disk(H3int h, int k):
     )
 
     ptr = create_ptr(n) # todo: return a "smart" pointer that knows its length?
-    check_for_error(
-        h3lib.gridDisk(h, k, ptr)
-    )
+
+    err = h3lib.gridDisk(h, k, ptr)
     mv = create_mv(ptr, n)
+    check_for_error(err)  # needs to be after, to ensure we clean up memory
 
     return mv
 
--- a/src/h3/_cy/cells.pyx
+++ b/src/h3/_cy/cells.pyx
@@ -166,6 +166,7 @@ cpdef H3int parent(H3int h, res=None) except 0:
     # (if you comment out the H3ResMismatchError in the error system file)
     # maybe need some way to indicate when extra info is expected (e.g. a message or unknown error code)
     # note/question: potential optimization benefit from not chaining function calls?
+    # update: maybe not since UnrecognizedH3ErrorCode
     if ex:
         msg = 'Invalid parent resolution {} for cell {}.'
         msg = msg.format(res, hex(h))
--- a/src/h3/_cy/cells.pyx
+++ b/src/h3/_cy/cells.pyx
@@ -12,7 +12,7 @@ from .util cimport (
 )
 
 from .error_system cimport check_for_error
-from .error_system import H3PentagonError, H3ResDomainError, H3ResMismatchError, H3Exception
+from .error_system import H3PentagonError, H3ResDomainError, H3ResMismatchError
 
 # todo: add notes about Cython exception handling
 
--- a/src/h3/_cy/cells.pyx
+++ b/src/h3/_cy/cells.pyx
@@ -222,6 +222,8 @@ cpdef H3int center_child(H3int h, res=None) except 0:
         msg = msg.format(res, hex(h))
         check_for_error(err, msg)
 
+    # todo: can we do this with something like a context manager?
+
     return child
 
 
@@ -401,15 +403,17 @@ cpdef get_faces(H3int h):
 
     check_cell(h)
 
-    # todo
-    err = h3lib.maxFaceCount(h, &n) #ignore error for now
+    check_for_error(
+        h3lib.maxFaceCount(h, &n)
+    )
 
     cdef int* ptr = <int*> stdlib.calloc(n, sizeof(int))
     if (n > 0) and (not ptr):
         raise MemoryError()
 
-    # todo
-    err = h3lib.getIcosahedronFaces(h, ptr) # handle error?
+    check_for_error(
+        h3lib.getIcosahedronFaces(h, ptr)
+    )
 
     faces = <int[:n]> ptr
     faces = {f for f in faces if f >= 0}
@@ -420,25 +424,22 @@ cpdef get_faces(H3int h):
 
 cpdef (int, int) experimental_h3_to_local_ij(H3int origin, H3int h) except *:
     cdef:
-        int flag
         h3lib.CoordIJ c
 
     check_cell(origin)
     check_cell(h)
 
-    # todo
-    flag = h3lib.cellToLocalIj(origin, h, 0, &c)
-
-    if flag != 0:
-        s = "Couldn't find local (i,j) between cells {} and {}."
-        s = s.format(hex(origin), hex(h))
-        raise H3PentagonError(s)
+    # todo: context manager for this idiom?
+    err = h3lib.cellToLocalIj(origin, h, 0, &c)
+    if err != 0:
+        msg = "Couldn't find local (i,j) between cells {} and {}."
+        msg = msg.format(hex(origin), hex(h))
+        check_for_error(err, msg)
 
     return c.i, c.j
 
 cpdef H3int experimental_local_ij_to_h3(H3int origin, int i, int j) except 0:
     cdef:
-        int flag
         h3lib.CoordIJ c
         H3int out
 
@@ -446,12 +447,12 @@ cpdef H3int experimental_local_ij_to_h3(H3int origin, int i, int j) except 0:
 
     c.i, c.j = i, j
 
-    # todo
-    flag = h3lib.localIjToCell(origin, &c, 0, &out)
+    # todo: context manager for this idiom?
+    err = h3lib.localIjToCell(origin, &c, 0, &out)
 
-    if flag != 0:
-        s = "Couldn't find cell at local ({},{}) from cell {}."
-        s = s.format(i, j, hex(origin))
-        raise H3PentagonError(s)
+    if err != 0:
+        msg = "Couldn't find cell at local ({},{}) from cell {}."
+        msg = msg.format(i, j, hex(origin))
+        check_for_error(err, msg)
 
     return out
--- a/src/h3/_cy/cells.pyx
+++ b/src/h3/_cy/cells.pyx
@@ -110,9 +110,7 @@ cpdef H3int[:] _ring_fallback(H3int h, int k):
     if dist_ptr is NULL:
         raise MemoryError() #todo: H3 memory error?
 
-    check_for_error(
-        h3lib.gridDiskDistances(h, k, ptr, dist_ptr)
-    )
+    err = h3lib.gridDiskDistances(h, k, ptr, dist_ptr)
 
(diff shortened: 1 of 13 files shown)


ajfriend:
After playing with this, I think it makes sense to land the changes I'm attempting in #261

src/h3/_cy/cells.pyx
```diff
@@ -125,25 +129,22 @@ cpdef H3int[:] ring(H3int h, int k):
...
-    # the memory. context manager would be better here, if we can figure out
-    # how to do that
-    mv = create_mv(ptr, n)
-
-    if err:
+    try:
+        check_for_error(
+            h3lib.gridRingUnsafe(h, k, ptr)
+        )
+    except H3ValueError:
```
isaacbrodsky, line 142: I would expect try/except to be relatively slow here compared to looking at the error value itself. That may not matter for this PR.

src/h3/_cy/error_system.pyx
```diff
@@ -0,0 +1,58 @@
...
+    pass
+
+cdef check_for_error(H3Error err):
+    """
+    todo: more descriptive message
+    todo: add error codes as property
+    todo: allow passing in extra args for the error message?
+    """
+
+    d = {
```
isaacbrodsky, line 36: Does this require allocating the error objects and dictionary each time a check is made?

src/h3/_cy/h3lib.pxd
```diff
@@ -2,9 +2,27 @@ from cpython cimport bool
 from libc.stdint cimport uint64_t, int64_t, uint32_t
 
 ctypedef uint64_t H3int
-ctypedef uint32_t H3Error
 ctypedef basestring H3str
 
+# todo: extern version of this?
+cdef enum H3Error:
```
isaacbrodsky, line 8: is there a way to inform the Cython compiler of the bit width of this enum?

src/h3/_cy/error_system.pyx
ajfriend: Maybe. But there's also a chance Cython optimizes that away. I know they do that with if-else chains by converting them to case statements when possible. We should definitely change this later for performance if that isn't the case tho. At this point, I feel like I'm still just prototyping the idea of what the error interface looks like.

src/h3/_cy/h3lib.pxd
ajfriend: Not sure yet. The confusion between the error type and the error codes was what led me to https://github.com/uber/h3/pull/613

Not sure what the most elegant solution is here, but definitely open to ideas!

ajfriend:
@isaacbrodsky, is it fair to think of the H3 error codes in a hierarchy? I'm now thinking the simplest thing is to have as close to a 1-1 mapping between the C error codes and the Python exceptions. We might also want to have some of the error classes inherit from the appropriate builtin Python exceptions, like `ValueError` and `MemoryError`.

Would this hierarchy make sense?

```
E_FAILED: H3Exception
    DNE: H3ValueError

        E_DOMAIN: H3DomainError
            E_LATLNG_DOMAIN: H3LatLngDomainError
            E_RES_DOMAIN: H3ResDomainError

        E_CELL_INVALID: H3CellInvalidError
        E_DIR_EDGE_INVALID: H3DirEdgeInvalidError
        E_UNDIR_EDGE_INVALID: H3UndirEdgeInvalidError
        E_VERTEX_INVALID: H3VertexInvalidError
        E_PENTAGON: H3PentagonError
        E_DUPLICATE_INPUT: H3DuplicateInputError
        E_NOT_NEIGHBORS: H3NotNeighborsError
        E_RES_MISMATCH: H3ResMismatchError
        E_OPTION_INVALID: H3OptionError

    DNE: H3MemoryError
        E_MEMORY: H3MemoryAllocationError
        E_MEMORY_BOUNDS: H3MemoryBoundsError
```
with something like
```
class H3Exception(Exception):
    pass

class H3ValueError(H3Exception, ValueError):
    pass

class H3MemoryError(H3Exception, MemoryError):
    pass
```

One weird thing is that we'd be creating two Python exceptions that don't map to a specific error code.

Q: Could we do something like rename `E_MEMORY` to `E_MEMORY_ALLOC`, and then use `E_MEMORY` as a parent of `E_MEMORY_ALLOC` and `E_MEMORY_BOUNDS`?

Q: Would it make sense to add an `E_VALUE`, that would basically be a parent error to all the non-memory-related H3 errors?

If we were to add those two, we could have a perfect 1-1 mapping between the C error codes and the H3-py exceptions.

isaacbrodsky:
I'm not sure it makes sense to think of the errors as a hierarchy. I can see certain errors being "programmer errors" (E_MEMORY_BOUNDS, E_OPTION_INVALID, E_FAILED) vs "data errors" (E_LATLNG_DOMAIN, E_CELL_INVALID, etc) but I am not sure a more complete hierarchy makes sense. I think it would be unique to the Python bindings in any case.

ajfriend:
> I'm not sure it makes sense to think of the errors as a hierarchy. I can see certain errors being "programmer errors" (E_MEMORY_BOUNDS, E_OPTION_INVALID, E_FAILED) vs "data errors" (E_LATLNG_DOMAIN, E_CELL_INVALID, etc) but I am not sure a more complete hierarchy makes sense.

Fair that this might be a Python thing. The Python error hierarchy is totally the thing that has me thinking about it.

Here's my hierarchy argument: The current error language already implies *some* hierarchy.

- `E_FAILED` is the parent of all other H3 errors. 
- `E_DOMAIN` is explained with: "Argument was outside of acceptable range (**when a more specific error code is not available**)"

So I don't think there's a debate around wether a hierarchy exists. It already does.

In this instance, it's just a question of wether it makes sense to branch off memory errors and value errors. That question could be answered differently across Python and C, but I think it'd be cleaner if they were the same.

src/h3/_cy/cells.pyx
ajfriend: I thought about it, and I think I've got a better approach with these latest changes. Let me know what you think!

src/h3/_cy/error_system.pyx
ajfriend: Also improved on this by moving the mapping outside of the function. LMKWYT!

ajfriend:
With all the changes above, this is basically a fresh PR, ready for some initial reviews :)

ajfriend:
Note, I did mess with some of the validation aggressiveness in this PR, but I should probably roll that back and defer any such changes to the discussion in https://github.com/uber/h3-py/issues/266

ajfriend:
# some todo items i might revisit in a second pass

- [x] tests for `h3_error_code`
- [ ] tests for unknown error codes
- [ ] think through `__init__.py`
- [ ] think through `.pxd`
- [ ] document error system file
- [ ] terminology for `H3GridNavigationError`?

ajfriend:
I think this is ready for review. I isn't perfect, but I think it counts as forward progress. I'm thinking I'd like to get this landed (if it looks like a reasonable start to everyone), and then build off of these changes in #261.

After landing #261, I'll come back and revisit the design here.

src/h3/_cy/cells.pyx
```diff
@@ -305,53 +301,51 @@ cpdef double mean_hex_area(int resolution, unit='km^2') except -1:
...
-        h3lib.H3Error err
         int64_t n
 
-    check_cell(start)
-    check_cell(end)
-
+    # todo: can we segfault here with invalid inputs?
+    # Can we trust the c library to validate the start/end cells?
+    # probably applies to all size/work pairs of functions...
     err = h3lib.gridPathCellsSize(start, end, &n)
```
isaacbrodsky, line 340: I'm not sure what this comments means

src/h3/_cy/cells.pyx
```diff
@@ -305,53 +301,51 @@ cpdef double mean_hex_area(int resolution, unit='km^2') except -1:
...
     else:
-        raise H3ValueError('Unknown unit: {}'.format(unit))
+        raise ValueError('Unknown unit: {}'.format(unit))
+
+    check_for_error(err)
 
     return area
 
 
+cdef couldnt_find_line(err, start, end):
```
isaacbrodsky, line 327: It might be best to avoid contractions in function names

src/h3/_cy/error_system.pyx
```diff
@@ -0,0 +1,235 @@
...
+    """
+    pass
+
+
+@contextmanager
+def the_error(obj):
+    """
+    Syntactic maple syrup for grouping exception definitions.
+    The associated `with` statement ends up as a not-half-bad
+    approximation to a valid sentence fragment.
```
isaacbrodsky, line 113: Nit as this does not effect anything about correctness, but this seems a little too cute for my taste. At a minimum I would add an underscore prefix to discourage importing this elsewhere.

src/h3/_cy/error_system.pyx
```diff
@@ -0,0 +1,235 @@
...
+# TODO: Move the helpers to util?
+# TODO: Unclear how/where to expose these functions. cdef/cpdef?
+
+cdef code_to_exception(H3Error err):
+    if err == E_SUCCESS:
+        return None
+    elif err in error_mapping:
+        return error_mapping[err]
+    else:
+        raise UnknownH3ErrorCode(err)
```
isaacbrodsky, line 220: It seems surprising that UnknownH3ErrorCode does not seem to derive from H3BaseException

src/h3/api/_api_template.py
```diff
@@ -254,7 +254,7 @@ def h3_distance(self, h1, h2):
         path between the cells in the graph formed by connecting
         adjacent cells.
 
-        This function will return an H3ValueError if the
+        This function will return an exception if the
```
isaacbrodsky, line 257: I assume this should read "raise an exception"?

src/h3/_cy/error_system.pyx
ajfriend: Makes sense. I'll change it.

src/h3/_cy/cells.pyx
ajfriend: Agreed.

src/h3/_cy/error_system.pyx
ajfriend: Definitely guilty of being cute. :)

I'm fine with whatever name suggestions you might have. I mostly just like the code construct for how clear and short
it makes the class hierarchy definitions:

```python
```

src/h3/_cy/cells.pyx
ajfriend: Is it possible to segfault with a pair of calls to `gridPathCellsSize`/`gridPathCells`? If this function were fuzzed, for example?

If we can't segfault and just get junk output, then for the Python bindings we have a choice to validate or not validate the inputs for the user.

However, if it is possible to segfault, we have no choice and we *must* validate the inputs well enough to guarantee that a segfault can't happen.

ajfriend:
The more I work on this PR, the less happy I am about mixing Python exceptions and Cython. I'll revisit the design (https://github.com/uber/h3-py/issues/267), but I'll land for now, just to move this along and unblock the improvements in #261 

@isaacbrodsky, let me know if you have any other concerns, and I'll can make an issue to make sure it gets addressed when I revisit.

PR #261, opened 2022-06-05
Use new H3MemoryManager system

ajfriend:


src/h3/_cy/memory.pyx
```diff
@@ -0,0 +1,115 @@
...
+    # todo: get rid of this?
+    # there's gotta be a better way to do this...
+    # create an empty cython.view.array?
+    cdef:
+        H3int a[1]
+
+    return (<H3int[:]>a)[:0]
+
+
+cdef class H3MemoryManager:
```
isaacbrodsky, line 59: I'd expect a memory manager to have broader responsibilities for managing allocations (have more verbs essentially); this seems to be more like a wrapped buffer or a block of managed memory?

ajfriend: Yeah, basically just an object to be responsible for deallocating the memory before that responsibility is handed off to a `memoryview` object.

Open to different names :)

src/h3/_cy/memory.pyx
```diff
@@ -0,0 +1,115 @@
...
+        # a[j-1] != 0
+        # i < j
+        # so we can swap! (actually, move a[j-1] -> a[i])
+        a[i] = a[j-1]
+        j -= 1
+
+    return i
+
+
+cdef H3int[:] empty_memory_view():
```
kylebarron, line 49: I know this is just copying an existing function, but can this have a docstring of what this function is used for? A memoryview isn't growable iirc, so not clear to me what the point of an empty memory view is.

src/h3/_cy/memory.pyx
kylebarron: Can we make this a context manager? Seems like a good fit unless for some reason a context manager is hard/slow in Cython

isaacbrodsky (approved):

src/h3/_cy/memory.pyx
ajfriend: I was also thinking about using a context manager here. I forget the specific examples, but it seemed like not every situation naturally fell into a context manager pattern.

An object that guarantees memory clean up would, at least, be an improvement over what we have now. I wouldn't be opposed to refactoring that into a context manager in the future.

Either way, I'll keep it in mind as I play with this. And if a context manager does end up working out, maybe I'll do that.

src/h3/_cy/memory.pyx
ajfriend: Ah, ok. I can try to add one.

The point of an empty memory view is that we could have situations where functions return an empty set.

IIRC, this was the cleanest way to do that. The syntax for creating one was pretty ugly otherwise, as far as I could figure it out, so I banished it to this function.

It'd be amazing if someone figured out a cleaner way to do this!

src/h3/_cy/memory.pxd
```diff
@@ -0,0 +1,19 @@
...
+
+
+cdef int[:] int_mv(size_t n)
+cdef H3int[:] empty_memory_view()
+
+"""
+todo: read: https://cython.readthedocs.io/en/latest/src/tutorial/pxd_files.html
+
+todo: should i be allocating with python memory functions?
+https://cython.readthedocs.io/en/latest/src/tutorial/memory_allocation.html
```
isaacbrodsky, line 18: I would expect this is preferred

src/h3/_cy/memory.pyx
```diff
@@ -0,0 +1,115 @@
...
+
+    return (<H3int[:]>a)[:0]
+
+
+cdef class H3MemoryManager:
+    def __cinit__(self, size_t n):
+        self.n = n
+        self.ptr = <H3int*> stdlib.calloc(self.n, sizeof(H3int))
+
+        if (self.n > 0) and (not self.ptr):
```
isaacbrodsky, line 64: nit: are the parenthesis needed?

src/h3/_cy/memory.pyx
isaacbrodsky, line 59: Please add a docstring as to what this class does. It's not immediately obvious it's specific to H3 indexes, although I think that's the case.

src/h3/_cy/memory.pyx
```diff
@@ -0,0 +1,115 @@
...
+        self.n = 0
+
+        return x
+
+    def __dealloc__(self):
+        # If the memory has been handed off to a memoryview, this pointer
+        # should be NULL, and deallocing on NULL is fine.
+        # If the pointer is *not* NULL, then this means the MemoryManager
+        # has is still responsible for the memory (it hasn't given the memory away to another object).
+        stdlib.free(self.ptr)
```
isaacbrodsky, line 95: Set self.ptr to NULL?

isaacbrodsky (approved):

src/h3/_cy/h3lib.pxd
```diff
@@ -10,7 +9,7 @@ cdef extern from "h3api.h":
     cdef int H3_VERSION_MINOR
     cdef int H3_VERSION_PATCH
 
-    ctypedef uint64_t H3Index
+    ctypedef uint64_t H3int "H3Index"
```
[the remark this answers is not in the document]
ajfriend: So that the Cython code only has one name for this concept. I'm using `H3int` here since the Python library deals with both integer and string representations of an `H3Index`, so and the names `H3int` and `H3str` help make it clear which one it is.

src/h3/_cy/memory.pxd
ajfriend: It turns out there are some interesting trade-offs we might consider around acquiring the GIL. Checkout the changes here: https://github.com/uber/h3-py/pull/261/commits/bf5fb8a079b46484d6470f24872e831f769b65e1

ajfriend: cc: @kylebarron

ajfriend:
Got history got weird. Moving these changes over to #268

PR #271, opened 2022-08-13
Remove alias functions

ajfriend:


ajfriend:
Given our target 4.0 launch, I'll just land this without review. Please let me know if you have any concerns.

PR #272, opened 2022-08-13
get_resolution, cell_to_parent, grid_distance

ajfriend:


docs/api_comparison.md
```diff
@@ -37,7 +37,7 @@ needs, based on speed and convenience.
 
 ```{tip}
 Note that the APIs are all 100% compatible, and it is easy to convert
-between them with functions like `h3_to_string` (links!) and `string_to_h3`.
+between them with functions like `int_to_string` (links!) and `string_to_int`.
```
isaacbrodsky, line 40: I don't think this matches the name in the core library. Is that expected?

ajfriend: Yeah. I think this is another case where specifying the `int` type is helpful. We also don't actually use the underlying C functions for these, and instead rely on Python's string/unicode support.

And I say this knowing there isn't going to be much appetite for it, but I'd honestly consider renaming the core library functions, as I think what `h3` means in this context isn't super clear to users.

ajfriend:
Remaining todos:

- area/length functions
- polyfill functions; probably want to consolidate the functions, and maybe reconsider the interface

docs/api_comparison.md
ajfriend: This also feels consistent to me with how we use `H3int` and `H3str` in the Python bindings.

ajfriend:
Hmmm. Getting some wheels/submodule errors. Let's try as "draft" and see if it works that way, at least.

PR #274, opened 2022-08-18
Get the makefile working on Linux

dfellis:
On many Linuxes, particularly Debian-derived ones, the Python 3 binary is `python3` instead of just `python` for weird legacy reasons, so added some logic to detect which python binary to use. Also switched all of the `env/bin/<foo>` calls to `./env/bin/<foo>` since the former only works if you have the current working directory in your `$PATH`, which is a major security no-no and I would be surprised if MacOS has that on by default (I think you guys must have accidentally added that to your shell setup).

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,21 +1,23 @@
 .PHONY: init clear rebuild purge test lint lab ipython
 
+PYTHON=$(shell command -v python || command -v python3)
+
 
 build-docs:
-	env/bin/jupyter-book build docs/ --warningiserror --keep-going --all
+	./env/bin/jupyter-book build docs/ --warningiserror --keep-going --all
 
 open:
 	open docs/_build/html/index.html
 
 init: purge
 	git submodule update --init
-	python -m venv env
-	env/bin/pip install --upgrade pip wheel setuptools
-	env/bin/pip install .[all]
-	env/bin/pip install -r requirements.in
+	$(PYTHON) -m venv env
+	./env/bin/pip install --upgrade pip wheel setuptools
+	./env/bin/pip install .[all]
+	./env/bin/pip install -r requirements.in
 
 clear:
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -1,21 +1,23 @@
 .PHONY: init clear rebuild purge test lint lab ipython
 
+PYTHON=$(shell command -v python || command -v python3)
```
dfellis, line 3: I know MacOS doesn't use `bash` anymore, so @ajfriend would you mind running `command -v python` and letting me know what it prints out for you? I can't remember if that's a bash built-in or a binary usable by any shell.

kylebarron: Works for me
```
bash-3.2$ command -v python
/Users/kbarron/.pyenv/shims/python
```

kylebarron (approved):

makefile
ajfriend: I get the same as Kyle, and on `zsh`.

ajfriend:
> On many Linuxes, particularly Debian-derived ones, the Python 3 binary is `python3` instead of just `python` for weird legacy reasons, so added some logic to detect which python binary to use.

I try to avoid touching system Python typically (even for setting up virtualenvs). So maybe implicit in this workflow is that you have `python` set up via something like `pyenv`: https://ajfriend.github.io/notes/mac_setup/python/

Either that or what you have here with `command -v python` should work. Not saying that we should do this, but if we were to assume everyone has `python` controlled by `pyenv`, then I don't think we'd need the `PYTHON=$(shell command -v python || command -v python3)`. On the other hand, that's probably a pain for folks who don't already have it set up and just want to make a quick change or test something, so this change is probably useful for the most folks.

ajfriend:
Tested the `makefile` locally, and it works for me!

PR #275, opened 2022-08-20
bump core lib to 4.0 rc5

ajfriend:


ajfriend:
Gonna land to move on to polyfill changes. Should be uncontroversial, and can always revisit before we merge to master.

PR #276, opened 2022-08-20
polygon_to_cells

ajfriend:


ajfriend:
Random thoughts: I'm starting to think through if it would be worth it to have a `Polygon` object in `h3-py` to simplify/clarify the `polyfill` (and related) functions. A GeoJSON-like (Multi)Polygon object is cursed with this confusing "list of lists of lists of lists of..." situation, mostly due to the optional (and maybe not that often used?) feature of holes.

Also, the current function in `h3-py` has its arguments in this awkward sequence:

```python
def polygon_to_cells(outer, int res, holes=None):
```

where the `res` parameter is inserted in the middle of defining the actual polygon (in the case that the polygon has holes). Alternatively, we could also do `polygon_to_cells(int res, outer, holes=None)` or `polygon_to_cells(int res, *loops)`, but having `res` first also seems odd to me.

Instead, we might have `h3.Polygon` objects with constructors like:

```python
class Polygon:
    def __init__(outer, holes=None): ...

class Polygon:
    def __init__(outer, *holes): ...

class Polygon:
    def __init__(*loops): ...
```

Then, we'd have something like:

```python
polygon_to_cells(poly1, res=9)
```

There's also currently an unfortunate mismatch between `polygon_to_cells` and `cells_to_multipolygon`. In contrast to a new `Polygon` object, I don't think we also need a new multipolygon object in python; it is simply a list of `Polygon`s. We'd then have something like

```python
def cells_to_multi_polygon(cells) -> list[h3.Polygon]: ...
```

The `Polygon` object could also have methods for polyfilling to cells, or for converting to and from different formats like GeoJSON.

Sketches of possible interfaces:

```python
h3.Polygon(
    outer = [(lat, lng), (lat, lng), ...],
    holes = [
        [(lat, lng), (lat, lng), ...],
        [(lat, lng), (lat, lng), ...],
    ],
)    

h3.Polygon(
    [(lat,lng), (lat, lng), ...],
    [(), ()],
    [(), ()],
).to_cells(res=9)

h3.Polygon(
    [(lat,lng), (lat, lng), ...],
).to_cells(res=9)
    
h3.polygon_to_cells(Polygon([(lat,lng), (lat, lng), ...]), res=9)

h3.Polygon([(lat,lng), (lat, lng), ...], holes=[[(), ()], [(),()]]).to_cells(res=9)
```

ajfriend:
Note that the `h3.Polygon` object isn't really anything more than a "list of lists" object. Its value is comes in from defining a specific object in the "list of lists of lists of lists..." data that we'd otherwise have to try to parse.

isaacbrodsky (approved):

ajfriend:
## TODO

- make sure all the docstrings are updated, especially the `Polygon` class, with some examples of usage
- update docs?

tests/test_polygon_class.py
```diff
@@ -0,0 +1,27 @@
...
+    a = '8928308280fffff'
+    b = h3.grid_ring(a, 5).pop()
+
+    cells1 = h3.grid_ring(b, 2) | {a}
+    cells2 = cells1 | {b}
+
+    poly1 = h3.cells_to_polygons(cells1)
+    poly2 = h3.cells_to_polygons(cells2)
+
+    # unfortunately output order is nondeterministic
```
ajfriend, line 14: Why is it that the output order of the polygons is nondeterministic? @isaacbrodsky @dfellis @nrabinowitz

src/h3/_cy/geo.pyx
```diff
@@ -54,35 +54,27 @@ cpdef (double, double) cell_to_latlng(H3int h) except *:
     return coord2deg(c)
 
 
-cdef h3lib.GeoLoop make_geoloop(geos, bool lnglat_order=False) except *:
```
nrabinowitz, line 57: You're no longer allowing `lng,lat` input? Or you've moved this handling elsewhere?

src/h3/_cy/geo.pyx
```diff
@@ -136,36 +124,40 @@ cdef class GeoPolygon:
...
-    The GeoJSON spec requires the right-hand rule, and a closed loop, but
-    this function will work with any input format.
+    The GeoJSON spec requires the right-hand rule and a closed loop, but
+    this function relaxes those constraints.
+
+    Unlike the GeoJson standard, the elements of the lat/lng pairs of each
+    ring are in lat/lng order, instead of lng/lat order.
+
+    We'll handle translation to different formats in the Python code,
+    rather than the Cython code.
```
nrabinowitz, line 144: I'm not sure this is wise for perf. In JS, we delay choosing lat/lng vs lng/lat until we're converting to C data formats, because that allows a single copy operation rather than multiple copy operations. With large/many polygons, copy operations could be expensive.

tests/test_polygon_class.py
nrabinowitz: Do you actually mean nondeterministic, or just undefined? There's nothing in the C lib that should be nondeterministic here, there's just no guarantee of order

ajfriend: Linking what you answered here: https://github.com/uber/h3-py/issues/277#issuecomment-1222662264

src/h3/_cy/geo.pyx
ajfriend: The input hasn't changed. It's still the same list of lat/lng pairs.

tests/test_polygon_class.py
ajfriend: It's actually non-deterministic. I'll get different output orders across different runs for the same input.

ajfriend: Running the same code (after restarting the interpreter):

```python
import h3

a = '8928308280fffff'
b = h3.grid_ring(a, 5).pop()

cells1 = h3.grid_ring(b, 2) | {a}
cells2 = cells1 | {b}

poly1 = h3.cells_to_polygons(cells1)
poly2 = h3.cells_to_polygons(cells2)

print(poly1)
print(poly2)
```

I got these different outputs:

```python
[<h3.Polygon |outer|=6, |holes|=()>, <h3.Polygon |outer|=30, |holes|=(18,)>]
[<h3.Polygon |outer|=6, |holes|=()>, <h3.Polygon |outer|=30, |holes|=(18,)>, <h3.Polygon |outer|=6, |holes|=()>]

[<h3.Polygon |outer|=30, |holes|=(18,)>, <h3.Polygon |outer|=6, |holes|=()>]
[<h3.Polygon |outer|=30, |holes|=(18,)>, <h3.Polygon |outer|=6, |holes|=()>, <h3.Polygon |outer|=6, |holes|=()>]

[<h3.Polygon |outer|=30, |holes|=(18,)>, <h3.Polygon |outer|=6, |holes|=()>]
[<h3.Polygon |outer|=6, |holes|=()>, <h3.Polygon |outer|=30, |holes|=(18,)>, <h3.Polygon |outer|=6, |holes|=()>]
```

ajfriend: Ahhh. I think it is because Python is not deterministic with set ordering. The following code seems to give consistent results:

```python
import h3

a = '8928308280fffff'
b = '89283082bc7ffff'
assert h3.grid_distance(a, b) == 5

cells1 = h3.grid_ring(b, 2) | {a}
cells2 = cells1 | {b}

cells1 = sorted(cells1)
cells2 = sorted(cells2)

poly1 = h3.cells_to_polygons(cells1)
poly2 = h3.cells_to_polygons(cells2)

print(poly1)
print(poly2)
```

PR #280, opened 2022-08-23
Prep for v4.0 beta 1 release

ajfriend:


isaacbrodsky (approved):

isaacbrodsky:
In the other bindings we have referred to the prelease version as a "release candidate", although this might have a different meaning than "beta".

ajfriend:
> In the other bindings we have referred to the prelease version as a "release candidate", although this might have a different meaning than "beta".

My understanding was that "release candidate" communicated an almost finalized stable state. As in, we could do a full release if no significant bugs were found in the release candidate. I wanted to communicate that this wasn't quite at that stage. Honestly, I think this is still in an "alpha" stage, but I think "beta" is the more widely recognized term.

PR #289, opened 2022-10-09
Fix broken example in README

timlinux:
When running through the simple example in the README, I get an error when trying to call ``latlng_to_cell``:

> AttributeError: module 'h3' has no attribute 'latlng_to_cell'

Looking at the python h3 API, this method is missing, but instead ``geo_to_h3`` produces an identical result to the example originally provided.

Versions in use:

image

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -53,7 +53,7 @@ conda install h3-py
 >>> import h3
 >>> lat, lng = 37.769377, -122.388903
 >>> resolution = 9
->>> h3.latlng_to_cell(lat, lng, resolution)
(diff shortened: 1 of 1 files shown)


isaacbrodsky:
The readme is referring to the prerelease v4.0.0 binding (v4.0.0-b1 I think is the version?) of H3, which renamed the geo_to_h3 function to latlng_to_cell.

isaacbrodsky:
@ajfriend I assume you don't intend to merge this

ajfriend:
Yes. Thanks!

PR #297, opened 2022-11-23
Prep for 4.0.0b2 release supporting py3.11

ajfriend:


kylebarron:
You also have to update the cibuildwheel version here:
https://github.com/uber/h3-py/blob/f25d24ff723af8170e3bc766b6a90a9c6052a8dd/.github/workflows/wheels.yml#L157

Looks like you need to update to at least 2.9.0 https://github.com/pypa/cibuildwheel/releases/tag/v2.9.0

PR #339, opened 2024-01-01
Replace _out_unordered and _out_ordered with _out_collection

ajfriend:
Change the APIs so that functions that would previously return sets of cells now return lists of cells.


tests/polyfill/test_h3.py
```diff
@@ -12,6 +12,16 @@ def __geo_interface__(self):
         return self.dictionary
 
 
+def same_set(a, b):
```
isaacbrodsky, line 15: it seems like this function gets duplicated a few times - maybe move it to a common test helper?

ajfriend: Done

ajfriend:
> You might want to mention in the respective doc strings that the ordering is not guaranteed from these functions.

Done

PR #359, opened 2024-04-14
update cibw

ajfriend:


ajfriend:
A potential way to clean up the build matrix: https://github.com/pypa/cibuildwheel/discussions/1261

ajfriend:
Seems like we don't lose anything by dropping `CIBW_ARCHS_MACOS: x86_64 arm64`

ajfriend:
Building wheels is by far the slowest part of h3-py development, so I'm interested in anything we can do that might speed it up.

We have one solution already, which is to avoid building wheels when a PR is in draft form, but that doesn't help for PRs like this one that are focused on building wheels :)

isaacbrodsky (approved):

PR #361, opened 2024-04-14
Fix docs upload

ajfriend:
- **try to fix deploy docs step**
- **should run just on a PR like this**


ajfriend:
Nope. That broke the website :/

ajfriend:
That got the website working again. :)

ajfriend:
And now it should only deploy to github pages when the PR is landed on `master`.

Issue #332, opened 2023-12-05
Change `H3Poly`/`H3MultiPoly` to `LatLngPoly`/`LatLngMultiPoly`


ajfriend:
Lat/Lng vs Lng/Lat order is one of the biggest sources for user bugs/questions/issues, especially when creating Polygon and MultiPolygon objects. Lng/Lat order is what's used in many other tools like `geopandas` and `GeoJSON`, but the H3 C library uses Lat/Lng order.

`h3-py` will stick with `Lat/Lng` order, but we can alleviate the mistakes caused from switching orders between tools by potentially making the ordering explicit in the name. An additional goal is to make it obvious that our polygon and multipolygon objects are different from `shapely` `Polygon` and `MultiPolygon` objects.

Some options:

1. `H3Poly` and `H3MultiPoly` (current; users need to read docstrings to know/recall ordering)
2. `LatLngPoly` and `LatLngMultiPoly` (explicit, but a bit long)
3. constructors like `H3Poly(outer, *holes, latlng=True)` (short name, "latlng choice" more prominent in the constructor signature than docstring)

I'm leaning towards `LatLngPoly` and `LatLngMultiPoly` but wanted to have a separate discussion here for future reference rather than having it hidden in the comments of #301.

I see it as a decision of how forcefully we want to remind the user of the ordering. In ordering of increasing forcefulness:

1. documentation or function docstrings
2. constructor signature
3. class name

PR #363, opened 2024-05-02
Prep for final v4.0 release!

ajfriend:

PR #364, opened 2024-05-04
Introduce LatLngPoly and LatLngMultiPoly

ajfriend:
Following from #332.

I'm leaving `H3Shape` instead of changing to just `Shape` because that seems too generic. And I think functions like `geo_to_shape` would be more confusing than `geo_to_h3shape`. Thoughts?

I'm also open to suggestions for a better name than `H3Shape`.


ajfriend:
Gotta drops builds like https://github.com/uber/h3-py/pull/363

docs/polygon_tutorial.ipynb
```diff
@@ -610,15 +610,15 @@
...
+    "`H3Shape` is an abstract class implemented by `LatLngPoly` and `LatLngMultiPoly`, each of which\n",
     "implement `__geo_interface__` and can be created from external \"Geo objects\":\n",
     "\n",
     "- `geo_to_h3shape()`\n",
     "- `h3shape_to_geo()`\n",
     "\n",
     "`H3Shape` objects can be converted to and from sets of H3 cells:\n",
     "\n",
-    "- `h3shape_to_cells()`\n",
+    "- `shape_to_cells()`\n",
```
dfellis, line 621: Was this accidentally changed?

src/h3/api/basic_int/__init__.py
```diff
@@ -432,49 +432,50 @@ def h3shape_to_cells(h3shape, res):
...
+        raise ValueError('Unrecognized type: ' + str(type(shape)))
 
     return _out_collection(mv)
 
 
 def cells_to_h3shape(cells, tight=True):
     """
-    Return an ``H3Shape`` describing the area covered by a collection of H3 cells.
-    Will return ``H3Poly`` or ``H3MultiPoly``.
+    Return an ``Shape`` describing the area covered by a collection of H3 cells.
```
dfellis, line 451: Similarly here. The docs say `Shape`, but the code is `H3Shape`

tests/polyfill/test_polyfill_ordering.py
```diff
@@ -1,4 +1,4 @@
-""" Test that `h3shape_to_cells` can take in polygon inputs
+""" Test that `shape_to_cells` can take in polygon inputs
```
dfellis, line 1: Same comment here about naming

docs/polygon_tutorial.ipynb
ajfriend: Good catch!

src/h3/api/basic_int/__init__.py
ajfriend: Good catch!

tests/polyfill/test_polyfill_ordering.py
ajfriend: Good catch!

src/h3/api/basic_int/__init__.py
```diff
@@ -395,14 +395,14 @@ def uncompact_cells(cells, res):
     return _out_collection(hu)
 
 
-def h3shape_to_cells(h3shape, res):
+def h3shape_to_cells(shape, res):
```
nrabinowitz, line 398: FWIW - I don't want to second-guess this, but I'm not sure how much this approach offers over just offering `polygon_to_cells` and making the caller break multi-polygons into polygons. The caller *still* has to handle this, by choosing `LatLngPoly` or `LatLngMultiPoly` when they convert their input data, but in this context you have to handle the both shapes in separate clauses and lose the clarity of sticking to the core library API. I'm not sure that trade-off is worthwhile for a slight ergonomic improvement when prepping input.

That said, I'm sure you've thought through this, so I'm okay with this API if you think it's worthwhile.

src/h3/api/basic_int/__init__.py
```diff
@@ -432,49 +432,50 @@ def h3shape_to_cells(h3shape, res):
...
         mv = _cy.polygon_to_cells(poly.outer, res, holes=poly.holes)
-    elif isinstance(h3shape, H3MultiPoly):
-        mpoly = h3shape
+    elif isinstance(shape, LatLngMultiPoly):
+        mpoly = shape
         mv = _cy.polygons_to_cells(mpoly.polys, res)
-    elif isinstance(h3shape, H3Shape):
-        raise ValueError('Unrecognized H3Shape: ' + str(h3shape))
+    elif isinstance(shape, H3Shape):
+        raise ValueError('Unrecognized Shape: ' + str(shape))
```
nrabinowitz, line 442: Nit: Probably would have kept this one `H3Shape`

src/h3/api/basic_int/__init__.py
```diff
@@ -432,49 +432,50 @@ def h3shape_to_cells(h3shape, res):
...
     else:
-        raise ValueError('Unrecognized type: ' + str(type(h3shape)))
+        raise ValueError('Unrecognized type: ' + str(type(shape)))
 
     return _out_collection(mv)
 
 
 def cells_to_h3shape(cells, tight=True):
```
nrabinowitz, line 449: Suggestion: Something I've been doing lately is `cells_to_h3shape(cells, *, tight=True)` to enforce kwargs for optional arguments - mostly to avoid `cells_to_h3shape(cells, False)` at the callsite, where the boolean argument is now opaque and confusing.

nrabinowitz (approved):
LGTM

dfellis (approved):

src/h3/api/basic_int/__init__.py
ajfriend: I like that better.

src/h3/api/basic_int/__init__.py
ajfriend: I was imagining this being helpful in the API because everything is symmetric and there are no special cases. We have three classes of objects: "geo objects", `H3Shape`s, and cells. For any two types, there's a function that translates between them, and every function has an obvious inverse. To me, that's a nice complete set of functionality, and there are no gotchas like "oh, right... for that case, I need to give it a list of polygons".

And I don't think it is true that the caller has to handle it by choosing either `LatLngPoly` or `LatLngMultiPoly`. I think the typical use case is actually that they call `geo_to_h3shape()` on a GeoPandas column (which are often a mix of Shapely Polygons and MultiPolygons). So think this aligns more with what Python folks expect from working with geopandas.

Also, we have a MultiPolygon class in the core API (its just kind of hard to use), so I think this actually still matches the C code.

isaacbrodsky (approved):

Issue #368, opened 2024-05-14
Add ability to handle (multi)polygons with Z-coordinate


ajfriend:
Some geometry objects come with a third coordinate (Z-coordinate). Functions like `geo_to_h3shape` break in this case, because they are only expecting 2 lat/lng coordinates.

See: https://github.com/uber/h3-py/issues/366

PR #371, opened 2024-05-18
Handle polygons with z-coordinate

ajfriend:
Make it so that we drop the Z coordinate (if it is present) when translating from a `__geo_interface__` object. `LatLngPoly` will still always expect just 2 coordinates in each point.

Closes https://github.com/uber/h3-py/issues/368


ajfriend:
FYI @fhk

dfellis (approved):

isaacbrodsky (approved):

PR #378, opened 2024-06-01
Trying pyproject.toml and scikit-build-core

ajfriend:
Trying for a [PEP 621](https://peps.python.org/pep-0621/) compatible setup using `pyproject.toml` instead of `setup.py`. Also trying to use `scikit-build-core`.

- https://learn.scientific-python.org/development/guides/packaging-compiled/
- https://scikit-build-core.readthedocs.io/en/latest/getting_started.html#cmake-file (Cython instructions)


ajfriend:
Currently running into an issue:

```
  CMake Error at src/h3/_cy/CMakeLists.txt:1 (find_package):
    No "FindCython.cmake" found in CMAKE_MODULE_PATH.


  CMake Warning (dev) at src/h3/_cy/CMakeLists.txt:1 (find_package):
    FindCython.cmake must either be part of this project itself, in this case
    adjust CMAKE_MODULE_PATH so that it points to the correct location inside
    its source tree.

    Or it must be installed by a package which has already been found via
    find_package().  In this case make sure that package has indeed been found
    and adjust CMAKE_MODULE_PATH to contain the location where that package has
    installed FindCython.cmake.  This must be a location provided by that
    package.  This error in general means that the buildsystem of this project
    is relying on a Find-module without ensuring that it is actually available.
```

Based on the comment "You’ll need to handle the generation of files by Cython directly at the moment. A helper (similar to scikit-build classic) might be added in the future." in https://scikit-build-core.readthedocs.io/en/latest/getting_started.html#cmake-file, I'm wondering if this is something from scikit-build classic that we need to remove and just do manually?

ajfriend:
Tests are running fine, but we need to figure out how to build the annotations again: https://github.com/scikit-build/cython-cmake/blob/main/src/cython_cmake/cmake/UseCython.cmake#L201-L202

ajfriend:
TODO: We were previously excluding unnecessary files in the sdist with `MANIFEST.in`. Those files are now being included, so we need to recover that functionality with this setup.

pyproject.toml
```diff
@@ -1,6 +1,90 @@
...
+    'dev_notes.md',
+    'makefile',
+]
+sdist.include = [
+    'src/h3lib/LICENSE',
+    'src/h3lib/README.md',
+    'src/h3lib/VERSION',
+    'src/h3lib/CMakeLists.txt',
+    'src/h3lib/cmake/*',
+    'src/h3lib/src/h3lib/*'
```
dfellis, line 89: Is this actually the path you want to include, or is this a typo?

ajfriend: That's the actual path! :)

dfellis (approved):

ajfriend:
Coverage decreased because I added the test files themselves to be included in the coverage report.

ajfriend:
> Coverage decreased because I added the test files themselves to be included in the coverage report.

And a good thing too, because it caught a bug!

Our Cython example is no longer working:


```
❯ env/bin/cythonize -i tests/test_cython/cython_example.pyx
/Users/aj/work/h3-py/tests/test_cython/cython_example.c:22015:13: warning: code will never be executed [-Wunreachable-code]
            goto bad;
            ^~~~~~~~
1 warning generated.
ld: warning: duplicate -rpath '/Users/aj/.pyenv/versions/3.12.3/lib' ignored
ld: warning: duplicate -rpath '/opt/homebrew/lib' ignored
error: could not create 'src/tests/test_cython/cython_example.cpython-312-darwin.so': No such file or directory
```

pyproject.toml
```diff
@@ -1,6 +1,93 @@
...
+[project.urls]
+Homepage = 'https://github.com/organization/package'
+Documentation = 'https://package.readthedocs.io/'
+'Bug Tracker' = 'https://github.com/organization/package/issues'
+Discussions = 'https://github.com/organization/package/discussions'
+Changelog = 'https://package.readthedocs.io/en/latest/changelog.html'
+
+
+[project.optional-dependencies]
+numpy = ['numpy<2']
```
isaacbrodsky, line 48: does this need to be limited to `numpy<2`?

isaacbrodsky (approved):

pyproject.toml
ajfriend: Nope! Dropped it.

ajfriend:
Everything should work, except for https://github.com/uber/h3-py/issues/396

I think we should merge this change to master and fix the issue from there.

Issue #323, opened 2023-07-02
Bind vertex functions


isaacbrodsky:
The functions at https://h3geo.org/docs/api/vertex/ are not in h3-py.

- [ ] [cellToVertex](https://h3geo.org/docs/api/vertex/#celltovertex)
- [ ] [cellToVertexes](https://h3geo.org/docs/api/vertex/#celltovertexes)
- [ ] [vertexToLatLng](https://h3geo.org/docs/api/vertex/#vertextolatlng)
- [ ] [isValidVertex](https://h3geo.org/docs/api/vertex/#isvalidvertex)

PR #388, opened 2024-08-24
Added Bindings for Vertex-Related Functions

Shiran-Yuan:
Added bindings and associated tests for the functions `cellToVertex`, `cellToVertexes`, `vertexToLatLng`, `isValidVertex`, and `check_vertex` per #323, ending the `h3-py` repo's longstanding disagreement with [official documentation](https://h3geo.org/docs/api/vertex).

Vertices are returned as integers by default in `cellToVertex` and `cellToVertexes`, but inputs to `vertexToLatLng` and `isValidVertex` can be either integers or strings (the two forms are equivalent via `int_to_str` and `str_to_int`).

Affected files:
- `src/h3/_cy/__init__.py`: added declarations
- `latlng.pxd` and `latlng.pyx`: wrote the functions
- `util.pxd` and `util.pyx`: added `check_vertex`
- `test_h3.py`: added tests
- `src/h3/api/basic_int/__init__.py` (and its aliases): added api access
- `h3lib.pxd`: added declarations for the original functions in `h3lib`
- `CHANGELOG.md` and `_version.py`: suggested version number `4.0.0b6`


ajfriend:
This is great. Thanks for the PR! I'll take a look.

CHANGELOG.md
```diff
@@ -16,7 +16,9 @@ avoid adding features or APIs which do not map onto the
 
 ## Unreleased
 
-- None
+## [4.0.0b6] - 2024-08-23
```
ajfriend, line 19: Let's just put this under `## Unreleased` for now, since we'll publish as a separate step.

src/h3/_cy/latlng.pyx
```diff
@@ -245,3 +246,46 @@ cpdef double great_circle_distance(
         raise ValueError('Unknown unit: {}'.format(unit))
 
     return d
+
+# todo: one might want to move the following to a separate vertex.pyx file
```
ajfriend, line 250: I'm supportive of moving this to a separate `vertex.pyx` file in this PR :)

src/h3/api/basic_int/__init__.py
```diff
@@ -996,3 +996,70 @@ def great_circle_distance(latlng1, latlng2, unit='km'):
...
+    -------
+    The vertex
+    """
+    return _cy.cell_to_vertex(_in_scalar(h), vertex_num)
+
+
+def cell_to_vertexes(h):
+    """
+    Return a list of vertices of an H3 cell. The list will always be of length 6.
+    If a pentagon is entered, the last element will be 0.
```
ajfriend, line 1021: In the Python world, I'd rather have us dynamically size the array: a hex should return an array of length 6, and a pentagon should return an array of length 5.

src/h3/api/basic_int/__init__.py
```diff
@@ -996,3 +996,70 @@ def great_circle_distance(latlng1, latlng2, unit='km'):
...
+
+    Parameters
+    ----------
+    h : H3Cell
+
+    Returns
+    -------
+    A list of vertices
```
ajfriend, line 1029: While I think we should have used "vertices" in H3, we decided to go with "vertexes". As a result, I think we should try to be consistent in the code and docs and only use "vertexes".

src/h3/api/basic_int/__init__.py
```diff
@@ -996,3 +996,70 @@ def great_circle_distance(latlng1, latlng2, unit='km'):
...
+    Returns
+    -------
+    lat : float
+        Latitude
+    lng : float
+        Longitude
+    """
+    if isinstance(v, str):
+        v = str_to_int(v)
+    return _cy.vertex_to_latlng(v)
```
ajfriend, line 1049: You need to use `_in_scalar()` so that this function will work as expected for the other APIs (`basic_int`, `basic_str`, etc.).

src/h3/api/basic_int/__init__.py
```diff
@@ -996,3 +996,70 @@ def great_circle_distance(latlng1, latlng2, unit='km'):
...
+    Validates an H3 vertex.
+
+    Returns
+    -------
+    bool
+    """
+    try:
+        if isinstance(v, str):
+            v = str_to_int(v)
+        return _cy.is_valid_vertex(v)
```
ajfriend, line 1063: Use `_in_scalar()` here as well. See how we do it in `is_valid_cell`, `is_valid_directed_edge`, and `is_pentagon`.

tests/test_h3.py
```diff
@@ -387,3 +387,62 @@ def test_grid_path_cells():
...
+def test_vertex_to_latlng():
+    latlng = h3.vertex_to_latlng('2114c3ffffffffff')
+    assert latlng == approx((24.945215618732814, -70.33904370008679))
+    latlng = h3.vertex_to_latlng(2455495337847029759)
+    assert latlng == approx((21.8146389946083, -57.23011196400803))
+
+
+def test_is_valid_vertex():
+    assert h3.is_valid_vertex('2114c3ffffffffff')
+    assert h3.is_valid_vertex(2455495337847029759)
```
ajfriend, line 446: We should expect this to fail for the (default) string API because a string is expected.

ajfriend:
Added some minor comments, but overall this looks great. Thanks!

tests/test_h3.py
```diff
@@ -387,3 +387,62 @@ def test_grid_path_cells():
...
+        2599918389178662911,
+        2527860795140734975,
+        2599931583318196223,
+    ]
+
+
+def test_vertex_to_latlng():
+    latlng = h3.vertex_to_latlng('2114c3ffffffffff')
+    assert latlng == approx((24.945215618732814, -70.33904370008679))
+    latlng = h3.vertex_to_latlng(2455495337847029759)
```
ajfriend, line 440: We don't expect any single API to be able to handle both `str` and `int` inputs. The current design expects them to select an API explicitly, like `h3.api.basic_int` or `h3.api.basic_str` (the default).

src/h3/_cy/latlng.pyx
Shiran-Yuan: Yeah but I thought that would be a larger change of structure... I wrote this comment in case there would be more vertex-related functions in the future. IMHO the current structure is good enough, but I can move it to a separate file as well if it's very beneficial :-)

src/h3/api/basic_int/__init__.py
Shiran-Yuan: Yeah I fixed array length to 6 only to align with the official docs... I'll change this.

src/h3/api/basic_int/__init__.py
Shiran-Yuan: OK! :-)

src/h3/api/basic_int/__init__.py
Shiran-Yuan: OK!

src/h3/api/basic_int/__init__.py
Shiran-Yuan: OK!

tests/test_h3.py
Shiran-Yuan: OK!

tests/test_h3.py
Shiran-Yuan: I see, I'll just leave the string tests, then.

CHANGELOG.md
Shiran-Yuan: OK!

src/h3/_cy/latlng.pyx
ajfriend: I appreciate you asking! I think your suggestion to move the vertex code to its own file is a good one. :)

Shiran-Yuan: Ok I'd do so.

src/h3/api/basic_int/__init__.py
```diff
@@ -996,3 +996,70 @@ def great_circle_distance(latlng1, latlng2, unit='km'):
...
+    ----------
+    h : H3Cell
+
+    Returns
+    -------
+    A list of vertexes
+    """
+    h = _in_scalar(h)
+    mv = _cy.cell_to_vertexes(h)
+    arr = [str_to_int(a) for a in _out_collection(mv)]
```
ajfriend, line 1034: Thanks for making adjustments, but this will still return `int`s for each API. We want the output to depend on the API:

- list of `int` for `h3.api.basic_int`
- list of `str` for `h3.api.basic_str`
- `numpy.array` for `h3.api.numpy_int`
- `memoryview` of `int` for `h3.api.memview_int`

To do that, the last operation should be `_out_collcation(mv)`. The `_remove_zeros()` function should help with this: https://github.com/uber/h3-py/blob/master/src/h3/_cy/memory.pyx#L113

Shiran-Yuan:
Update: All previously proposed comments by @ajfriend have now been resolved! :-)

src/h3/api/basic_int/__init__.py
isaacbrodsky: Is this line still needed? It looks as if you can `return _out_collection(mv)`?

ajfriend:
While I'm trying to fix the codedov issue, @Shiran-Yuan, would you mind also adding these functions to the docs? It should just involve creating a new section for these vertex functions in https://github.com/uber/h3-py/blob/master/docs/api_quick.md

Shiran-Yuan:
> While I'm trying to fix the codedov issue, @Shiran-Yuan, would you mind also adding these functions to the docs? It should just involve creating a new section for these vertex functions in https://github.com/uber/h3-py/blob/master/docs/api_quick.md

I see, no problem, I'll go add that.

ajfriend (approved):
This looks good to me! Thanks for the PR! I think we can land without the codecov since it isn't clear when that will get resolved. I'll leave this up for a bit to see if anyone else has comments.

PR #389, opened 2024-08-25
try new codecov token

ajfriend:


ajfriend:
Hmm. seems like it isn't just us: https://github.com/codecov/codecov-action/issues/1508

PR #409, opened 2024-10-05
try testing numpy <2 and >=2

ajfriend:


ajfriend:
Abandon in favor of #410

PR #408, opened 2024-10-05
Support `str`-like values for str_to_int

jongbinjung:
Since `numpy>=2` (?), `str` values are not native `str` but `numpy.str_`. 

Without the explicit conversion to `str`, the test fails with:

```
================================================== FAILURES ===================================================
_______________________________________________ test_numpy_str ________________________________________________

    def test_numpy_str():
        expected = [
            617700169957507071,
            617700169957769215,
            617700169958031359,
            617700169958293503,
            617700169961177087,
            617700169964847103,
            617700169965109247,
        ]
        cells = np.array([h3.int_to_str(h) for h in expected])
>       got = [h3.str_to_int(c) for c in cells]

tests/test_apis/test_numpy_int.py:50:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_apis/test_numpy_int.py:50: in <listcomp>
    got = [h3.str_to_int(c) for c in cells]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

h = np.str_('89283082803ffff')

    def str_to_int(h):
        """
        Converts a hexadecimal string to an H3 64-bit integer index.

        Parameters
        ----------
        h : str
            Hexadecimal string like ``'89754e64993ffff'``

        Returns
        -------
        int
            Unsigned 64-bit integer
        """
        # return _cy.str_to_int(str(h))
>       return _cy.str_to_int(h)
E       TypeError: Argument 'h' has incorrect type (expected str, got numpy.str_)

env/lib/python3.11/site-packages/h3/api/numpy_int/__init__.py:57: TypeError
``` 

* [Test at failing state](https://github.com/uber/h3-py/actions/runs/11194484081/job/31121593037)

PR #411, opened 2024-10-05
Allow string subtypes as H3str

jongbinjung:
A better resolution of #408 using [auto encoding/decoding](https://docs.cython.org/en/stable/src/tutorial/strings.html#auto-encoding-and-decoding).

See #412 for [broken tests](https://github.com/uber/h3-py/actions/runs/11195853327/job/31124011714?pr=412).


ajfriend (approved):
Looks good to me. Let's just get one more stamp before landing.

dfellis (approved):

isaacbrodsky:
LGTM

Issue #393, opened 2024-09-03
Find Codecov alternative


ajfriend:
Codecov seems to be broken, as in https://github.com/uber/h3-py/pull/389

ajfriend:
A possible alternative: https://hynek.me/articles/ditch-codecov-python/

ajfriend:
Weird. Codecov is working again? https://github.com/uber/h3-py/pull/394

ajfriend:
As noted in https://github.com/uber/h3-py/pull/408, this seems to be a known issue with Codecov and it doesn't look like it will be resolved any time soon:

- https://github.com/codecov/codecov-action/issues/29
- https://github.com/pypa/twine/issues/658

Let's find an alternative solution.

jongbinjung:
Going to mark closed, by #413 

We can re-open if we decide otherwise.

PR #413, opened 2024-10-05
Replace codecov with local lcov

jongbinjung:
Potential solution to #393 

The reporting isn't as pretty, and you don't get the full report. But it gets the job done without sending anything outside of the Github Actions instance.

Setup to use local coverage reporting for pull requests, and codecov for push to master. 

Seems like once pushed to master, codecov pulls the proper secrets from `uber/h3-py` (e.g., see [this](https://github.com/uber/h3-py/actions/runs/11195132272) run on `master` vs [this](https://github.com/uber/h3-py/actions/runs/11194900791) failed run on a fork for the PR )

We _could_ just use lcov for everything, but then we'd have to give up the badge on `readme.md`.

### DIFF
--- a/.github/workflows/coverage-lint-pr.yml
+++ b/.github/workflows/coverage-lint-pr.yml
@@ -1,8 +1,8 @@
-name: coverage-lint
+name: pr-coverage-lint
 
 on:
-  push:
-    branches: [master]
+  pull_request:
+    branches: ['*']
 
 jobs:
   tests:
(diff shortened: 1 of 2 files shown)


ajfriend:
This is great! And this should be sufficient, especially when we figure out the Cython tests to get 100% coverage. Let's try this out for a while, while we mull over the importance of the badge. :)

ajfriend (approved):

Issue #383, opened 2024-06-22
Single-source version with dynamic = ["version"]


ajfriend:
See if we can use `dynamic = ["version"]` in `pyproject.toml` to single-source the package version. I tried this in #378, but couldn't get it working.

- https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

PR #415, opened 2024-10-11
Use dynamic versioning to single-source version from git tags

jongbinjung:
Does this address #383 ? 

Example local install:

```
$ env/bin/python
>>> import h3
>>> h3.__version__
'4.1.0b4.dev0+g6d9a5fc.d20241011'
```

and then with a tag:

```
$ git tag v4.1.0b4
$ env/bin/pip install .
...
$ env/bin/python
>>> import h3; h3.__version__
'4.1.0b4'
```

Currently using the [default versioning scheme](https://setuptools-scm.readthedocs.io/en/latest/usage/#default-versioning-scheme); but configurable. 

Wasn't sure what didn't work last time (in #378)

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4.1.1
         with:
+          # Git-based versioning needs the full history
+          fetch-depth: 0
           submodules: recursive
 
       - uses: actions/setup-python@v5.0.0
--- a/.github/workflows/coverage-lint-pr.yml
+++ b/.github/workflows/coverage-lint-pr.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4.1.1
         with:
+          # Git-based versioning needs the full history
+          fetch-depth: 0
           submodules: recursive
 
       - uses: actions/setup-python@v5.0.0
--- a/.github/workflows/coverage-lint.yml
+++ b/.github/workflows/coverage-lint.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4.1.1
         with:
+          # Git-based versioning needs the full history
+          fetch-depth: 0
           submodules: recursive
 
       - uses: actions/setup-python@v5.0.0
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4.1.1
         with:
+          # Git-based versioning needs the full history
+          fetch-depth: 0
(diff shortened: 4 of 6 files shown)


jongbinjung:
Not sure what these failing unit tests are ....

jongbinjung:
> Not sure what these failing unit tests are ....

OK, figured out it was because Github Action workflows only do a shallow fetch; but the "single source" of version with the PR becomes the git history ... 

I've tried a couple things to just pull the git tags, but they all failed. Checked out a some other repositories that deploy via Github actions ([prefect](https://github.com/PrefectHQ/prefect/blob/main/.github/workflows/python-package.yaml#L23-L24), [pandas](https://github.com/pandas-dev/pandas/blob/main/.github/workflows/wheels.yml#L53)), and looks like they all do a full fetch.

ajfriend:
Oh, this is slightly different from what I was thinking. I simply wanted to define the version in either `pyproject.toml` or `h3.__version__`, and have the other read it. I had accomplished (one direction of) it with https://github.com/uber/h3-py/blob/121314a6212b7687d8c1df4192358faf2fb87a49/src/h3/_version.py#L1-L3 and I guess I forgot to close the ticket. 

This is cool tho. I do really like the `'4.1.0b4.dev0+g6d9a5fc.d20241011'` hash for development. However, I do also like having the version explicitly in the code somewhere as opposed to a tag. (But I suppose a tag is a *true* single source of truth...) Is that an option?

But maybe I just need to get used to this :) What do other folks think? @isaacbrodsky @nrabinowitz @dfellis

jongbinjung:
> Oh, this is slightly different from what I was thinking. I simply wanted to define the version in either `pyproject.toml` or `h3.__version__`, and have the other read it. I had accomplished (one direction of) it with
> 
> https://github.com/uber/h3-py/blob/121314a6212b7687d8c1df4192358faf2fb87a49/src/h3/_version.py#L1-L3
> 
> and I guess I forgot to close the ticket.
> This is cool tho. I do really like the `'4.1.0b4.dev0+g6d9a5fc.d20241011'` hash for development. However, I do also like having the version explicitly in the code somewhere as opposed to a tag. (But I suppose a tag is a _true_ single source of truth...) Is that an option?
> 
> But maybe I just need to get used to this :) What do other folks think? @isaacbrodsky @nrabinowitz @dfellis

I see. The only way I can think of for having a `tag` be source of truth and also have the "current" version be available somewhere explicitly would be to have some janky job run at tag creation, that writes some change to `master`? ... but maybe I just lack imagination.

The main benefit of using tags, imo, is that you can cut a release without having that one odd PR just to bump the version in-code 😅 (e.g., via Github's Release UI)—but I can also see how that might not be necessarily desirable.

ajfriend:
Closing for now, but maybe we'll revisit in the future.

Issue #337, opened 2024-01-01
Try ruff


ajfriend:

jongbinjung:
What happened? You didn't like?

ajfriend:
Closed with #428

PR #416, opened 2024-10-11
Replace flake8 and pylint with ruff

jongbinjung:
In case we're still interested in #337 

Tried my best to accommodate existing code style. 

Also added review dog to github actions, so that it's easier to see broken things in the PR directly


ajfriend (changes requested):
I do like adding ruff! And I can mess with the style options later :)

I'm not familiar with review dog. Despite the nice onomatopoeic pairing, can we split out the review dog change to a different PR?

tests/test_apis/test_basic_int.py
```diff
@@ -9,7 +8,7 @@ def test_int_output():
     lng = -122.418307270836
 
     assert h3.latlng_to_cell(lat, lng, 9) == 617700169958293503
-    assert h3.latlng_to_cell(lat, lng, 9) == 0x8928308280fffff
```
ajfriend, line 12: Can we change this rule? We use lower case 'f' pretty much everywhere else.

jongbinjung: I don't think we can.

The linter has a bunch of configurations, but the fixer—which is the thing that actually makes these changes—is much more opinionated. We _could_ just skip the fixer, but then the dev workflow would involve running the linter and manually addressing any issues that arise ... 

Given this specific occurrence just happens once in a test, we could also just mare this one line to be skipped?

ajfriend: Sure. Let's just skip it for now, and maybe complain upstream to see if we can get a fixer option for this :)

docs/polygon_tutorial.ipynb
```diff
@@ -34,41 +34,48 @@
...
     "    plot_df(df, ax=ax)\n",
     "\n",
+    "\n",
     "def plot_cells(cells, ax=None):\n",
     "    shape = h3.cells_to_h3shape(cells)\n",
     "    plot_shape(shape, ax=ax)\n",
     "\n",
+    "\n",
     "def plot_shape_and_cells(shape, res=9):\n",
-    "    fig, axs = plt.subplots(1,2, figsize=(10,5), sharex=True, sharey=True)\n",
```
ajfriend, line 71: 🙄

docs/polygon_tutorial.ipynb
```diff
@@ -429,7 +433,13 @@
     "poly1 = h3.LatLngPoly([(37.804, -122.412), (37.778, -122.507), (37.733, -122.501)])\n",
     "poly2 = h3.LatLngPoly(\n",
     "    [(37.803, -122.408), (37.736, -122.491), (37.738, -122.380), (37.787, -122.39)],\n",
-    "    [(37.760, -122.441), (37.772, -122.427), (37.773, -122.404), (37.758, -122.401), (37.745, -122.428)]\n",
```
ajfriend, line 432: I was about to complain about this, but its inconsistent with what I just said above. 😬

src/h3/api/basic_int/__init__.py
```diff
@@ -3,17 +3,16 @@
...
 )
-
 from ._convert import (
-    _in_scalar,
-    _out_scalar,
     _in_collection,
+    _in_scalar,
     _out_collection,
+    _out_scalar,
 )
```
ajfriend, line 16: 🙄

tests/polyfill/test_h3.py
```diff
@@ -360,8 +355,12 @@ def test_cells_to_h3shape_non_contiguous():
...
     # Six hexagons in a ring around a hole
     cells = [
-        '892830828c7ffff', '892830828d7ffff', '8928308289bffff',
-        '89283082813ffff', '8928308288fffff', '89283082883ffff',
+        '892830828c7ffff',
+        '892830828d7ffff',
+        '8928308289bffff',
+        '89283082813ffff',
+        '8928308288fffff',
+        '89283082883ffff',
```
ajfriend, line 363: This is better :)

tests/polyfill/test_h3.py
```diff
@@ -400,7 +398,7 @@ def test_multipoly_checks():
 def test_3d_geo():
     loop = lnglat_open()
 
-    loop2d = [(lng, lat)      for lng, lat in loop]
+    loop2d = [(lng, lat) for lng, lat in loop]
```
ajfriend, line 401: Worse

tests/polyfill/test_polyfill.py
```diff
@@ -1,42 +1,35 @@
...
     # big center chunk of the US in lat/lng order
     outer = [
         [42.68, -110.61],
         [32.17, -109.02],
-        [31.57,  -94.26],
-        [42.94,  -89.38],
-        [42.68, -110.61]
+        [31.57, -94.26],
+        [42.94, -89.38],
+        [42.68, -110.61],
```
ajfriend, line 16: Worse. But i do like that it added the trailing comma :)

tests/polyfill/test_polyfill.py
```diff
@@ -1,42 +1,35 @@
...
+        [39.77, -105.07],
     ]
 
-    hole2 = [
-        [41.37, -98.61],
-        [40.04, -91.80],
-        [42.32, -91.80],
-        [41.37, -98.61]
-    ]
+    hole2 = [[41.37, -98.61], [40.04, -91.80], [42.32, -91.80], [41.37, -98.61]]
```
ajfriend, line 27: Worse

tests/polyfill/test_polyfill.py
```diff
@@ -68,7 +61,7 @@ def test_h3shape_to_cells():
         '832b1afffffffff',
         '832b1efffffffff',
         '832ba9fffffffff',
-        '832badfffffffff'
+        '832badfffffffff',
```
ajfriend, line 64: Nice

ajfriend:
I like the idea of `ruff`, but I don't know if I like these particular rules. Can we start with a minimal `ruff` config that  changes almost nothing, and we can add rules from there? 

Maybe just start with the most important rules? 
- double quotes to single quotes
- trailing commas

:P

docs/polygon_tutorial.ipynb
```diff
@@ -92,11 +99,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "outer = [\n",
-    "    (37.804, -122.412),\n",
-    "    (37.778, -122.507),\n",
-    "    (37.733, -122.501)\n",
-    "]\n",
+    "outer = [(37.804, -122.412), (37.778, -122.507), (37.733, -122.501)]\n",
```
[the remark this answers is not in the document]
jongbinjung: The way a lot of these auto formatters work is that you can add a trailing comma to the last element to force it to have one element per line.

tests/polyfill/test_h3.py
jongbinjung: Why? Is there any benefit of having the `for` in the two loops aligned?

tests/polyfill/test_polyfill.py
jongbinjung: Yes. Adding the trailing comma will auto fmt this to be element-per-line.

jongbinjung (commented):
The formatter (cf. linter) which does the automatic formatting isn't really configurable. The _linter_ is, but if we're just going to use the linter and not autofmt, might as well stick with `flake8`.

But, I think the issues you pointed out are:
* Per-element line breaks for collections that are within line length: This can be forced with trailing commas
* Aligning on the decimal
* Whitespace preference

The latter two—while not really supported—seem to only come up in tests, where there's a lot of personal "touches" to describe/build test data. So, we could just have the formatter skip unit tests?

pyproject.toml
```diff
@@ -46,10 +46,10 @@ Changelog = 'https://package.readthedocs.io/en/latest/changelog.html'
 
 [project.optional-dependencies]
 numpy = ['numpy']
-test = ['pytest', 'pytest-cov', 'flake8', 'pylint', 'numpy']
+test = ['pytest', 'pytest-cov', 'ruff', 'numpy']
 all = [
+    'h3[test]',
```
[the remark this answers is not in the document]
ajfriend: Stole this in #425

ajfriend:
Ruff has started to bring me some joy with #428

ajfriend:
Thanks for this! I stole some ideas and used them in https://github.com/uber/h3-py/pull/428

PR #432, opened 2024-12-15
upgrade to h3 v4.2.0

isaacbrodsky:


src/h3/api/basic_int/__init__.py
```diff
@@ -477,6 +477,10 @@ def h3shape_to_cells(h3shape, res):
     h3shape : ``H3Shape``
     res : int
         Resolution of the output cells
+    flags : int
+        Containment mode flags
```
isaacbrodsky, line 481: `flags` seems like it needs an enum

isaacbrodsky:
Coverage seems to be getting 403 which I don't think is related to this change.

tests/test_cells_and_edges.py
```diff
@@ -289,11 +289,11 @@ def test_average_hexagon_area():
...
-        0: 1107.712591000,
-        1: 418.676005500,
-        2: 158.244655800,
-        9: 0.174375668,
-        15: 0.000509713,
+        0: 1281.256011,
+        1: 483.0568391,
+        2: 182.5129565,
+        9: 0.200786148,
+        15: 0.000584169,
```
ajfriend, line 296: This is a pretty big change. Do we know what happened here?

isaacbrodsky: This appears to be reflecting https://github.com/uber/h3/pull/749

ajfriend: OK, I guess I thought we had this in a while ago.

ajfriend:
I was looking for docs on the containment modes and realized that we don't have it documented. As part of this effort, I think we should do:

- https://github.com/uber/h3/issues/956
- https://github.com/uber/h3/issues/957

src/h3/api/basic_int/__init__.py
```diff
@@ -467,7 +467,7 @@ def uncompact_cells(cells, res):
     return _out_collection(hu)
 
 
-def h3shape_to_cells(h3shape, res):
+def h3shape_to_cells(h3shape, res, flags=0, experimental=False):
```
ajfriend, line 470: I'm not sure if this should be the final form of the API. For example, we might not use integers for the flags, and we may drop the `experimental` option.

Could we leave `h3shape_to_cells` unchanged, and just create a `h3shape_to_cells_experimental` where we're free to change the API?

dfellis (approved):

ajfriend:
This looks great. thanks!

PR #434, opened 2024-12-29
Simplify coverage and lint actions

ajfriend:
For https://github.com/uber/h3-py/issues/433

PR #435, opened 2024-12-29
Simplify coverage and lint actions v2

ajfriend:
Separate tests between Python and Cython APIs; see the `tests/readme.md`. We do this because we are currently not at 100% coverage for the Cython API (note: which isn't currently externally supported), but we are at 100% coverage for the Python API.

We were previously using a coverage service, but ran into issues often enough that it seems easier to just setup a test requiring 100% coverage on the Python API. In the future, we'll also require 100% on the Cython API.

This PR also adds some tests to get back to 100% after adding the experimental polyfill algorithm in #432


ajfriend:
Subsumes #434

isaacbrodsky (approved):

PR #470, opened 2025-11-18
Bump h3lib to v4.4.1 and add new functions/errors

ajfriend:
- Update `h3lib` to v4.4.1. (#470)
    - Add new error codes:
        - `H3IndexInvalidError`
        - `H3BaseCellDomainError`
        - `H3DigitDomainError`
        - `H3DeletedDigitError`
    - Add new functions:
        - `is_valid_index`
        - `get_index_digit`
        - `construct_cell`
        - `deconstruct_cell` (note that this function doesn't currently exist in C library)

### DIFF
--- a/docs/api_quick.md
+++ b/docs/api_quick.md
@@ -51,8 +51,8 @@ and should be generally aligned with the
    get_resolution
    get_base_cell_number
    get_index_digit
-   deconstruct_cell
    construct_cell
+   deconstruct_cell
 ```
 
 ## Geographic coordinates


src/h3/_cy/cells.pyx
```diff
@@ -49,6 +58,33 @@ cpdef int get_resolution(H3int h) except -1:
 
     return h3lib.getResolution(h)
 
+cpdef int get_index_digit(H3int h, int res) except -1:
+    cdef:
+        int digit
+
+    check_cell(h)
```
isaacbrodsky: should this be check_index rather?

PR #493, opened 2026-06-18
Fix wheels for python 3.14 free thread

### LINKED ISSUE #488: Support for free-threaded Python 3.14
Many python packages are starting to enable the free-threaded python versions. It would be nice if this could be considered for this package since there is at least one downstream package (timezonefinder) working on this direction https://github.com/jannikmi/timezonefinder/issues/364. 

I did some investigation by installing h3-py alongside [pytest-run-parallel](https://github.com/jannikmi/timezonefinder/issues/364#:~:text=pytest%2Drun%2Dparallel) and found no issue (while working on this one must set the environment variable `PYTHON_GIL=0`). 

Although this result only indicates that maybe this library can be deployed to thread-free python, it is also possible that tests are not sufficiently extensive and didn't touch cases where it would fail in a thread-free python run.

aureliobarbosa:
This PR annotates .pyx cython files to avoid GIL being re-enabled on cpython 3.14 free thread during the execution of packages including h3-py with this python version.

Previously I did run pytest with [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel) script and got green, so, for independent parallel process is expected to run without problems. 

Fix #488

Available for questions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ avoid adding features or APIs which do not map onto the
 [H3 core API](https://uber.github.io/h3/#/documentation/api-reference/).
 
 ## Unreleased
+- Fix wheels for Python 3.14.6 free-threading
 
 ## [4.5.0] - 2026-05-29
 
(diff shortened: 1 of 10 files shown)


dfellis (approved):

jongbinjung:
Thanks!

PR #496, opened 2026-07-26
test: clean up string formatting TODO in test_parent_err

dyrpsf:
### Description
This is a quick cleanup PR to resolve a lingering `todo` comment I noticed while looking through the test suite. 

In `test_parent_err` inside `tests/test_lib/test_cells_and_edges.py`, there was a comment to revisit the "weird formatting stuff" used to generate the hex error message. I replaced the clunky `hex(h3.str_to_int(h))` conversion with a standard, readable Python f-string: `f'Invalid parent resolution -1 for cell 0x{h}.'`

This makes the test much more idiomatic and removes the TODO without changing any of the underlying test logic.

### Maintainers
cc: @dfellis @ajfriend @isaacbrodsky - Just a small housekeeping fix for whenever you have the time to review!

### DIFF
--- a/tests/test_lib/test_cells_and_edges.py
+++ b/tests/test_lib/test_cells_and_edges.py
@@ -131,7 +131,7 @@ def test_parent_err():
     except Exception as e:
         msg = str(e)
 
-    expected = f'Invalid parent resolution -1 for cell 0x{h}.'
+    expected = f'Invalid parent resolution -1 for cell {h}.'
 
     assert msg == expected
 
--- a/tests/test_lib/test_cells_and_edges.py
+++ b/tests/test_lib/test_cells_and_edges.py
@@ -131,7 +131,7 @@ def test_parent_err():
     except Exception as e:
         msg = str(e)
 
-    expected = f'Invalid parent resolution -1 for cell {h}.'
+    expected = f'Invalid parent resolution -1 for cell 0x{h}.'
 
     assert msg == expected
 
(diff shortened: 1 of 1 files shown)


dfellis (approved):

tests/test_lib/test_cells_and_edges.py
```diff
@@ -131,9 +131,7 @@ def test_parent_err():
     except Exception as e:
         msg = str(e)
 
-    # todo: revist this weird formatting stuff
-    expected = 'Invalid parent resolution -1 for cell {}.'
-    expected = expected.format(hex(h3.str_to_int(h)))
+    expected = f'Invalid parent resolution -1 for cell 0x{h}.'
```
isaacbrodsky:
```suggestion
    expected = f'Invalid parent resolution -1 for cell {h}.'
```
I don't think we need to prefix the cell, since hexadecimal cell IDs are usually printed as is

isaacbrodsky (approved):

dyrpsf:
@isaacbrodsky I've applied your suggestion to remove the 0x prefix! Let me know if you need to re-approve the CI workflows for this new commit, or if it is fully good to merge.

dyrpsf:
@isaacbrodsky It looks like the CI tests failed after applying the suggestion! I dug into it, and it turns out the underlying library's exception message actually does include the `0x` prefix (which is why the original test was using Python's `hex()` function). I've added the `0x` back to the f-string so the assertion passes again. It should be fully green and ready to go now!

dfellis (approved):

isaacbrodsky (approved):

dyrpsf:
Grateful for the merge—thank you! I’m here if any additional changes are needed.

